PR #26, opened 2014-04-07
Support for RubySaml 0.8.x added

esbeka:
Fix to support RubySaml 0.8.x.
Fixes issue: PracticallyGreen/omniauth-saml#25

The multiple value support as mentioned by @borgand should still be implemented.
Would it be possible to create a separate issue for that? As the lack of RubySaml 0.8 support blocks the upgrade of Nokogiri to name just one.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,20 +1,21 @@
 PATH
   remote: .
   specs:
-    omniauth-saml (1.1.0)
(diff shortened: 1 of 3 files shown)


bpedro:
Merged https://github.com/PracticallyGreen/omniauth-saml/pull/35 which does the same.

PR #27, opened 2014-06-11
Update ruby-saml gem and naming

johnmetta:
Earlier versions of the ruby-saml gem lock other gems that have been updated and are needed by us, maybe others.


bpedro:
Merged https://github.com/PracticallyGreen/omniauth-saml/pull/35 which adds support for ruby-saml 0.8.x

Issue #28, opened 2014-08-01
Support additional SAML token fields in info hash


alim:
It would be useful to support additional SAML token fields in the info hash that is returned during the callback phase. In particular, the additional information to be included in the info hash could include:
- issuer 
- session_index
- session_not_on_or_after
- response.session_expires_at
- not_before
- not_on_or_after

These values are available from the ruby-saml gem [Response class](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/response.rb). I would be happy to supply a pull request with the corresponding code and test updates.

md5:
The whole `Response` object is now available via `extra['response_object']`, available since version 1.7.0. See #107.

PR #29, opened 2014-08-01
Added support for additional SAML token fields to the info hash in the callback phase.

alim:
This pull request is for Issue #28 on the PracticallyGreen/omniauth-saml repository. It adds an expanded number of info hash fields during the callback phase. The fields are pulled from the ruby-saml Response class. The updated fields include:

* issuer 
* session_index
* session_not_on_or_after
* response.session_expires_at
* not_before
* not_on_or_after

An additional test case was added to validate the new fields.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,8 +1,8 @@
 PATH
   remote: .
   specs:
-    omniauth-saml (1.1.0)
-      omniauth (~> 1.1)
+    omniauth-saml (1.2.0)
+      omniauth (~> 1.2)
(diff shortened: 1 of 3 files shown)


md5:
I think a better approach here would be to add the entire `OneLogin::SAML::Response` as the value of a `:saml_response` field under `extra` (during the callback phase).

supernova32:
Closing this pull request since it has had no activity in more than 6 months. @alim if you would like to see this functionality added, please create a new pull request based from the current master, and take @md5's comment into consideration.

PR #36, opened 2015-01-27
Return multi values for attributes and support Windows Identity Information

maveonair:
Thank you very much for this great Gem which allows us to integrate a SAML authentication very simple!

For my use case, I had just the issue that I have to use a Windows Identity SAML provider, you name it, that returns for a certain key multiple values as for example memberships groups. For that reason I updated Ruby-Saml to version 0.9 which has this features. Unfortunately, it leads to the fact that each attribute in attributes is returned as a Array, e.g. ```:name => ["John Doe"]``` as you can see in the updated spec. 

Moreover, I want to use this Gem for several different applications and I extended the information hash to fetch some details with the "stupid" Windows Identity Schema XML Namespace names to get this information. I am aware that you maybe not interested in this namespace cr*** but nevertheless it allows me to use this awesome Gem for SAML authentication in a Windows Identity environment.


maveonair:
Custom attributes can be used after the following PR is merged: https://github.com/omniauth/omniauth-saml/pull/72/commits Therefore I close this PR.

md5:
@maveonair #72 doesn't support multi-valued attributes. I think that would be a good addition, but it isn't covered by the current PR.

maveonair:
@md5 In my case I just wanted to use the windows attributes because the others were not applicable.Therefore the #72 solved my problem...

PR #43, opened 2015-05-13
Upgrade ruby-saml dependency

thiagoxvo:
I've just upgrade ruby-saml dependency and updated the tests to pass on.

The upgrade was necessary due a bug we found in our stack and we were not able to parse an offline saml schema:
- Torquebox 3.2.1 
- JRuby 
- Nokogiri 1.6.1 (patched - we can't update because of Torquebox)

So the only solution was upgrade some of gems we are dependent and omniauth-saml was one of then.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,20 +1,20 @@
 PATH
   remote: .
   specs:
-    omniauth-saml (1.3.0)
+    omniauth-saml (1.3.1)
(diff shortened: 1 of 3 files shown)


bpedro:
Closing because https://github.com/PracticallyGreen/omniauth-saml/pull/46 obsoletes this.

Thanks!

PR #49, opened 2015-09-03
Support for Office 365 (Azure AD) SAML response

patope:
FYI. I got Azure AD SAML authentication to work with gitlab after this change.

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -104,10 +104,10 @@ module OmniAuth
 
       info do
         {
-          :name  => @attributes[:name],
-          :email => @attributes[:email] || @attributes[:mail],
-          :first_name => @attributes[:first_name] || @attributes[:firstname] || @attributes[:firstName],
-          :last_name => @attributes[:last_name] || @attributes[:lastname] || @attributes[:lastName]
+          :name  => @attributes['http://schemas.microsoft.com/identity/claims/displayname'] || @attributes[:name],
+          :email => @attributes['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name'] || @attributes[:email] || @attributes[:mail],
+          :first_name => @attributes['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname'] || @attributes[:first_name] || @attributes[:firstname] || @attributes[:firstName],
+          :last_name => @attributes['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname'] || @attributes[:last_name] || @attributes[:lastname] || @attributes[:lastName]
         }
       end
 


patope:
here is example of o365 saml response 
https://gist.github.com/patope/a55f8b73d0a12c27f1c3

lib/omniauth/strategies/saml.rb
```diff
@@ -104,10 +104,10 @@ def other_phase
 
       info do
         {
-          :name  => @attributes[:name],
-          :email => @attributes[:email] || @attributes[:mail],
-          :first_name => @attributes[:first_name] || @attributes[:firstname] || @attributes[:firstName],
-          :last_name => @attributes[:last_name] || @attributes[:lastname] || @attributes[:lastName]
+          :name  => @attributes['http://schemas.microsoft.com/identity/claims/displayname'] || @attributes[:name],
+          :email => @attributes['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name'] || @attributes[:email] || @attributes[:mail],
```
md5: This should probably be `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress`, not `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name`.

md5:
I think this PR makes sense only if it limits itself to claims from the `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/` namespace, not the `http://schemas.microsoft.com/identity/claims/` namespace.

BTW, it's currently possible to integrate with Azure AD without this PR by simply defining the attributes that `omniauth-saml` currently expects for the application in question. See this screenshot for an example:

screen shot 2015-10-04 at 3 56 56 pm

I understand that that could be tedious to do for every application, but it also doesn't make a lot of sense to me to change this gem in ways that are only necessary to support Azure AD.

md5:
Note, the first five attribute mappings are the ones configured by Azure AD by default. The one that's obscured by the tooltip is `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name`.

lib/omniauth/strategies/saml.rb
patope: See https://gist.github.com/patope/a55f8b73d0a12c27f1c3 there is no attribute "emailaddress". Guess "name" is O365 login name, which is same as email.

patope:
What Azure AD you are using? Premium? I'm using one from O365 for Small Business Premium that comes with really basic features and i couldn't find where saml attributes can be reconfigured.

md5:
I'm on the "Pay-As-You-Go" plan.

When I go to my "Default Directory", I click on "Applications", then on the application name. Then, I click "Attributes" on the top bar, where I can add attributes with the "Add User Attribute" button.

patope:
On application page i only see 

- Dashboard
- Users
- Configure
- Owners

 I think O365 bundled Azure AD is quite limited in functionality and mainly meant for ready made integrations. Anyway, i have this configuration in production (gitlab) and works fine for us. This pull request more like example of fix than clean fix. We are using this same integration in Redmine. In Redmine attributes can be mapped in saml configuration. (Well, is has own issues too https://github.com/chrodriguez/redmine_omniauth_saml/issues/6). 

Maybe fix like this https://github.com/flowdock/omniauth-saml/commit/abdd336a5fb274a5c600d8d7c5aa46823de5dcb6 would be better..

md5:
@patope Yep. I think that would work well.

It looks like Office 365 is using the user's `mail` attribute as the `NameID` while the default for my Pay-As-You-Go Azure account is to use `userprincipalname`.

md5:
I believe this was functionality is now provided by #72. Please reopen if you think otherwise.

PR #50, opened 2015-09-07
Adding "settings" to Response Class at initialization to handle signing verification

gmontard:
According to Ruby SAML Doc in order to have signing validated you have to initialize the call OneLogin::RubySaml::Response using RubySaml Settings and do not this later.

Thanks.

### DIFF
--- a/lib/omniauth-saml/version.rb
+++ b/lib/omniauth-saml/version.rb
@@ -1,5 +1,5 @@
 module OmniAuth
   module SAML
-    VERSION = '1.4.1'
+    VERSION = '1.4.2'
   end
 end
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -6,6 +6,8 @@ module OmniAuth
     class SAML
       include OmniAuth::Strategy
 
+      OTHER_REQUEST_OPTIONS = [:skip_conditions, :allowed_clock_drift, :matches_request_id, :skip_subject_confirmation].freeze
+
       option :name_identifier_format, nil
       option :idp_sso_target_url_runtime_params, {}
       option :request_attributes, [
@@ -47,7 +49,15 @@ module OmniAuth
         end
 
         settings = OneLogin::RubySaml::Settings.new(options)
-        response = OneLogin::RubySaml::Response.new(request.params['SAMLResponse'], settings: settings)
+        # filter options to select only extra parameters
+        opts = options.select {|k,_| OTHER_REQUEST_OPTIONS.include?(k.to_sym)}
+        # symbolize keys without activeSupport/symbolize_keys (ruby-saml use symbols)
+        opts =
+          opts.inject({}) do |new_hash, (key, value)|
+            new_hash[key.to_sym] = value
+            new_hash
+          end
+        response = OneLogin::RubySaml::Response.new(request.params['SAMLResponse'], opts.merge(settings: settings))
         response.attributes['fingerprint'] = options.idp_cert_fingerprint
 
         @name_id = response.name_id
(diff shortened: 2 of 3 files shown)


md5:
@gmontard Could you provide a link to the docs you're talking about?

I just took a look at the `ruby-saml` code and I don't see any reason it would not validate the signature with the `omniauth-saml` code is it now stands. In particular, [these lines](https://github.com/PracticallyGreen/omniauth-saml/blob/7b9b18136d49674c7bcbed5f2d48e52003616ca6/lib/omniauth/strategies/saml.rb#L60-L62) will validate signed elements among other things and raise a validation error for an invalid response.

md5:
According to the comments over in #58, it looks like the problem is actually with decrypting responses, not with validating signed elements.

That part does indeed only happen if the `:settings` object is passed to `initialize` (cf. https://github.com/onelogin/ruby-saml/blob/5ecd6a01a8e82892a03eb1c13e4a394b97636bd3/lib/onelogin/ruby-saml/response.rb#L58-L60).

gmontard:
@md5 Sorry for the late response.
You are right this as to do with the encryption but it doesn't change anything at the end.

Here are what the doc says:
> but **if the SAMLResponse contains an encrypted assertion, you need to provide the settings in the initialize method in order to be able to obtain the decrypted assertion**, using the service provider private key in order to decrypt. If you don't know what expect, use always the first proposed way (always set the settings on the initialize method).

Thanks.

alex:
FWIW, I can confirm that we're unable to handle encrypted assertions without this PR. (If there's anything I can do to help get it landed, please let me know)

alex:
@bpedro Is there anything I could do to help get this merged?

colmben:
+1 - Response encryption doesn't work as the response initialisation process needs the SP private_key setting. It looks like the current code was actually trying to do it both ways (pass the settings into the Response.new and also set the settings afterwards), but the Response.new bit just passes in the straight options hash and doesn't set it up with the required settings: key. So that ", options" bit in the Response.new is pointless, it does nothing. It would be really great if we could get this into master asap.

zakkak:
:+1:

ilikepi:
:+1:

tomilaine:
:+1:

bosskovic:
:+1:

yasseur-vodeclic:
Hey guys, we've just committed some changes on @gmontard fork to fix some options that were not passed to ruby-saml. Hope this will be merged!

gmontard:
:+1:

md5:
@yasseur-vodeclic I took at look at this PR and the changes to `ruby-saml` and it's unclear to me why those properties were not added to the `OneLogin::RubySaml::Settings` class instead of being passed as separate parameters to `OneLogin::RubySaml::Response.new`. It seems like it would be worth making that change to `ruby-saml` if possible since it would simplify the calling code for all users of that gem.

supernova32:
I agree with @md5, but I'm not sure how much work that would be.

@yasseur-vodeclic could you fix the conflicts in the meantime?

yasseur-vodeclic:
@md5 I'll check as soon as possible the impact of adding OneLogin::RubySaml::Settings to Response :). Yes @supernova32 I'll fix the conflicts

md5:
@gmontard or @yasseur-vodeclic Do you mind rebasing this branch to make it mergable? I think there's consensus to add this change to the next release.

md5:
I've opened https://github.com/omniauth/omniauth-saml/pull/82 to replace this PR

yasseur-vodeclic:
Hi!
Thanks for the new PR, i was going to notice you here, but you were too fast :D

Issue #45, opened 2015-06-29
name of email attribute in SAML callback response isn't expected by saml.rb


carlost:
https://github.com/PracticallyGreen/omniauth-saml/blob/master/lib/omniauth/strategies/saml.rb#L93

It appears that saml.rb expects the email attribute to have a name of "email" or "mail".  The SAML IDP we are authenticating against returns "emailAddress".  I don't know if that is non-standard but we can't do anything about it.

I monkey patched the class to get this working in our environment.

Would you be open to adding optional configuration parameters to map the attributes to the values expected in the info block?

```ruby
option :email_attribute_name, :email
option :first_name_attribute_name, :first_name
option :last_name_attribute_name, :last_name
option :name_attribute_name, :name
...
info do
        {
          :name  => @attributes[options[:name_attribute_name]],
          :email => @attributes[options[:email_attribute_name]],
          :first_name => @attributes[options[:first_name_attribute_name]],
          :last_name => @attributes[options[:last_name_attribute_name]]
        }
end
```
The sample above may break some existing installs because currently the SAML runs through a few potential names for a few of these attributes.  So to replicate that behavior we could do the following:

```ruby
option :email_attribute_name, :email
option :first_name_attribute_name, :first_name
option :last_name_attribute_name, :last_name
option :name_attribute_name, :name
...
info do
        {
          :name  => @attributes[options[:name_attribute_name]],
          :email => @attributes[options[:email_attribute_name]] || @attributes[:mail],
          :first_name => @attributes[options[:first_name_attribute_name]]  || @attributes[:firstname] || @attributes[:firstName],
          :last_name => @attributes[options[:last_name_attribute_name]] || @attributes[:lastname] || @attributes[:lastName]
        }
end
```
I will work on the PR if you are receptive to this approach.

md5:
This could be fixed by #51.

PR #51, opened 2015-09-16
Support custom attributes

OsQu:
Attribute statements might not always have a Name="firstName" attribute, and modifying the idp to support can be an exhausting task (e.g. ADFS).

Now one can give `attribute_statements` option, which contains a mapping from statements to an info hash. For example the default mapping is following:

      option :attribute_statements, {
        name: ["name"],
        email: ["email", "mail"],
      }

This we can support whatever weird configurations the other end might have, without hard-coding them directly to the source.

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -15,6 +15,12 @@ module OmniAuth
         { name: 'last_name', name_format: 'urn:oasis:names:tc:SAML:2.0:attrname-format:basic', friendly_name: 'Family name' }
       ]
       option :attribute_service_name, 'Required attributes'
+      option :attribute_statements, {
+        name: ["name"],
+        email: ["email", "mail"],
+        first_name: ["first_name", "firstname", "firstName"],
+        last_name: ["last_name", "lastname", "lastName"]
+      }
 
       def request_phase
         options[:assertion_consumer_service_url] ||= callback_url
@@ -103,15 +109,23 @@ module OmniAuth
       uid { @name_id }
 
       info do
-        {
-          :name  => @attributes[:name],
-          :email => @attributes[:email] || @attributes[:mail],
-          :first_name => @attributes[:first_name] || @attributes[:firstname] || @attributes[:firstName],
-          :last_name => @attributes[:last_name] || @attributes[:lastname] || @attributes[:lastName]
-        }
+        found_attributes = options.attribute_statements.map do |key, values|
+          attribute = find_attribute_by(values)
+          [key, attribute]
+        end
+
+        found_attributes.to_h
       end
 
       extra { { :raw_info => @attributes } }
+
+      def find_attribute_by(keys)
+        keys.each do |key|
+          return @attributes[key] if @attributes[key]
+        end
+
+        nil
+      end
     end
   end
 end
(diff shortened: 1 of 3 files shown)


md5:
Looks like this covers the use case in #49 and part of the use case in #36.

It seems like it might make sense to extend this mapping to cover the multi-valued use case in #36 as well.

I'm thinking something like this:

```ruby
option :attribute_statements, {
    name: ["name"],
    email: ["email", "mail"],
    last_name: ["last_name", "lastname", "lastName"],
    groups: { attributes: ['http://schemas.xmlsoap.org/claims/Group'], multi: true }
}
```

The logic would be that if an array is seen, it is assumed to be an `attributes` list and `multi` is `false`. If a `Hash` is seen, then the `attributes` and `multi` entries are checked, with `attributes` being a required entry and `multi` defaulting to `false`.

md5:
Note: I just meant that as an example; I don't think that `groups` should have a mapping by default.

md5:
cc/ @maveonair @patope

md5:
This PR would also address #45.

lib/omniauth/strategies/saml.rb
```diff
@@ -103,15 +109,23 @@ def other_phase
...
-          :email => @attributes[:email] || @attributes[:mail],
-          :first_name => @attributes[:first_name] || @attributes[:firstname] || @attributes[:firstName],
-          :last_name => @attributes[:last_name] || @attributes[:lastname] || @attributes[:lastName]
-        }
+        found_attributes = options.attribute_statements.map do |key, values|
+          attribute = find_attribute_by(values)
+          [key, attribute]
+        end
+
+        found_attributes.to_h
```
ilikepi: Array#to_h is not present in Ruby 2.0 and earlier.  Using `Hash[found_attributes]` instead seems to work under Ruby 2.0 and is still supported under Ruby 1.9.3 and 2.2 according to the docs.

supernova32:
Having such an option would really remove the headache of having to modify the idP in order to match what is expected. @OsQu can you address @ilikepi's comment? The gem should also be compatible with older versions of Ruby.

ilikepi:
I opened a PR with my suggested fix against the branch on which this PR is based.

supernova32:
thanks @ilikepi. I'd really like to get this merged. Let's hope @OsQu accepts your PR soon.

It would also be very nice to include an example of how to use the custom attributes in the README as a guide to all users.

OsQu:
Great job people and sorry for the longish radio silence!

Unfortunately I don't have access to Flowdock's organization anymore and I've moved on to different adventures already. I can however fork the repo under my own account and apply @ilikepi's changes and issue a new PR to omniauth-saml if you think this is a good change. I can also update README.

Thoughts?

md5:
@OsQu I think in that case it would work just as well to create a new PR from @ilikepi's branch: https://github.com/PracticallyGreen/omniauth-saml/compare/master...ilikepi:custom_attributes_compat_fix

I'd personally like to see the Travis CI support in #68 merged before we tackle any more PRs, but we have to figure out whether Ruby 1.8.7 should be supported before that's mergeable.

supernova32:
Closing in favor of #72

PR #56, opened 2015-10-02
Make a copy of @options when this strategy instance is dup'd

### LINKED ISSUE #54: Fallback ACS URL gets reused across requests under Devise
I just spent a bunch of time debugging an issue using `omniauth-saml` with Devise. We are using the "fallback" functionality introduced in #15 to dynamically set our `:assertion_consumer_service_url` to the Omniauth `callback_url`.

The issue we ran into was that the `:assertion_consumer_service_url` was being calculated and cached based on the `callback_url` of the first request to be handled. This was then being used for subsequent requests, even if the hostname for the request was not the same, resulting in the user being sent to the wrong hostname after logging in with the IdP.

I wasn't able to determine whether this is a bug in `rails`, `devise`, `omniauth`, or `omniauth-saml`, but I was able to work around it by monkey-patching this strategy like so:

```ruby
class OmniAuth::Strategies::SAML
  alias_method :original_request_phase, :request_phase
  def request_phase
    options[:assertion_consumer_service_url] = nil
    original_request_phase
  end
end
```

During my debugging, I was able to determine that a different instance of `OmniAuth::Strategies::SAML` was being used for each call to `request_phase`, but that those instances all shared the same instance of `OmniAuth::Strategy::Options`. However, when I added an `initialize` method to my monkey-patched version of `OmniAuth::Strategies::SAML`, I did not see it getting called more than once.

I'm still going to try to reproduce this in a minimal test case, but I wanted to open this issue in case anyone else has observed this problem.

A simple fix in the case of `omniauth-saml` might be to avoid any code that mutates `options`, replacing it with code that makes a copy of `options` first before making any changes it needs (e.g. the defaulting of `options[:assertion_consumer_service_url]`.

md5:
Since `OmniAuth::Strategy#call` does `dup` before calling `call!`, this allows the `options` to be mutated without affecting subsequent requests.

Fixes #54

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -112,6 +112,15 @@ module OmniAuth
       end
 
       extra { { :raw_info => @attributes } }
+
+      protected
+
+        # Make a copy of @options when this strategy instance is dup'd in OmniAuth::Strategy#call.
+        # This allows the options to be mutated without affecting subsequent requests.
+        def initialize_copy(orig)
+          super
+          @options = @options.deep_dup
+        end
     end
   end
 end


md5:
Sorry for the @-tag @options! You must get it more than I get people accidentally tagging me with @md5 :grinning:

md5:
@bpedro It would be great to get some feedback on this PR when you have a moment. :+1:

md5:
I just realized that this patch won't work outside of a Rails environment because `#deep_dup` is a Rails addition. Outside of Rails, calling `@options.deep_dup` will return `nil`, since Hashie will interpret it as `@options['deep_dup']`.

I'm going to work on a patch for `omniauth` itself that works outside of Rails.

md5:
I'm going to close this PR. I think the right thing to do in `omniauth-saml` is to simply avoid mutating `@options` at all by doing local `dup` within the `request_phase` and `callback_phase` methods. I'll open a new PR with this change instead.

md5:
The replacement PR is at https://github.com/PracticallyGreen/omniauth-saml/pull/63

jturkel:
@md5 - If `@options` is an instance of `Hashie::Mash`, then won't calling `@options.dup` do a deep copy?

```
>> source = Hashie::Mash.new(a: {b: 1})
#<Hashie::Mash a=#<Hashie::Mash b=1>>
>> copy = source.dup
#<Hashie::Mash a=#<Hashie::Mash b=1>>
>> copy[:a][:c] = 2
2
>> copy.inspect
"#<Hashie::Mash a=#<Hashie::Mash b=1 c=2>>"
>> source.inspect
"#<Hashie::Mash a=#<Hashie::Mash b=1>>"
```

md5:
Good catch @jturkel! I wasn't aware of that behavior of `Hashie::Mash`.

So now I guess I'm back to opening a PR against `omniauth`...

md5:
@jturkel I just opened https://github.com/intridea/omniauth/pull/829

Please take a look! :+1:

jturkel:
Thanks @md5! Looks great.

md5:
Unfortunately, it's failing on Ruby 1.8.7 and JRuby in 1.8 mode, which `omniauth` still supports. Given that, I'm not sure how to proceed. I'd rather not have to override `dup` instead, but that's what would be needed to support Ruby 1.8 (see [this blog post](http://www.jonathanleighton.com/articles/2011/initialize_clone-initialize_dup-and-initialize_copy-in-ruby/) for a nice overview).

PR #58, opened 2015-10-09
Preload SAML settings before processing assertion

paultag:
In case we need `:private_key`, :settings should be set.


alex:
To expand, if the payload is encrypted, `initialize` needs to have a settings.

paultag:
Changeset #50 looks the same.

@bpedro - any thoughts on #50?

Issue #55, opened 2015-09-30
Check for success never happens because of earlier check for name_id


djfpaagman:
Right now there is no check on the StatusCode, which [can returns multiple failures][1]. Instead the first error that is triggered is "SAML response missing 'name_id'' while there is actually might be more going on underneath, making it hard to debug.

[ruby-saml does check it][2] in its validation, but since the check for `name_id` happens earlier, you will never see it. I'm not really sure what the right solution would be, I think moving up the `valid?` call could do the trick but not sure of the consequences of that.

I'm happy to explore the possibilities but checking here first.

[1]: https://msdn.microsoft.com/en-us/library/hh269642.aspx
[2]: https://github.com/onelogin/ruby-saml/blob/a331dac9fd3bcfa716c599dbe4e961e4638e2d14/lib/onelogin/ruby-saml/response.rb#L264

djfpaagman:
Ping

md5:
:+1:

ilikepi:
:+1:

supernova32:
@djfpaagman thanks for this, I merged #60

PR #60, opened 2015-10-15
Call validation earlier to get real error instead of 'response missing name_id'

djfpaagman:
I proposed this change in #55. This will call `valid?` earlier, so that the real error checking in `ruby-saml` happens before the blank check on `@name_id`, resulting in more meaningful errors in case something is wrong with the SAML response.


supernova32:
The explanation given on #55 makes me wonder why this was not merged sooner. Got 2 :+1: on the issue, and I also like this, so merged!

PR #63, opened 2015-11-12
Add specs for ACS fallback URL behavior

### LINKED ISSUE #54: Fallback ACS URL gets reused across requests under Devise

md5:
This PR updates `Omniauth::Strategies::SAML` to avoid mutating the `options` hash during `request_phase` and `callback_phase`. This mutation was the underlying cause of the failure reported in #54.

This PR is an alternative to the fix proposed in #56, since `deep_dup` is Rails-specific.

Let me know if you want me to drop the `Gemfile.lock` changes.

Also, I targeted this at the `master` branch instead of the `develop` branch because the latter branch is now behind.

Fixes #54

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     omniauth-saml (1.4.2)
-      omniauth (~> 1.1)
+      omniauth (~> 1.3)
       ruby-saml (~> 1.1, >= 1.1.1)
 
 GEM
--- a/omniauth-saml.gemspec
+++ b/omniauth-saml.gemspec
@@ -11,7 +11,7 @@ Gem::Specification.new do |gem|
   gem.email         = 'rajiv@alum.mit.edu'
   gem.homepage      = 'https://github.com/PracticallyGreen/omniauth-saml'
 
-  gem.add_runtime_dependency 'omniauth', '~> 1.1'
+  gem.add_runtime_dependency 'omniauth', '~> 1.3'
   gem.add_runtime_dependency 'ruby-saml', '~> 1.1', '>= 1.1.1'
 
   gem.add_development_dependency 'rspec', '~> 2.8'
--- a/spec/omniauth/strategies/saml_spec.rb
+++ b/spec/omniauth/strategies/saml_spec.rb
@@ -17,7 +17,7 @@ describe OmniAuth::Strategies::SAML, :type => :strategy do
   let(:saml_options) do
     {
       :assertion_consumer_service_url     => "http://localhost:3000/auth/saml/callback",
-      :idp_sso_target_url                 => "https://idp.sso.target_url/signon/29490",
+      :idp_sso_target_url                 => "https://idp.sso.example.com/signon/29490",
       :idp_cert_fingerprint               => "C1:59:74:2B:E8:0C:6C:A9:41:0F:6E:83:F6:D1:52:25:45:58:89:FB",
       :idp_sso_target_url_runtime_params  => {:original_param_key => :mapped_param_key},
       :name_identifier_format             => "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress",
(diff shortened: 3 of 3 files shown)


md5:
Following a tip from @jturkel that `Hashie::Mash#dup` does a deep copy, I opened a PR for `omniauth` over at https://github.com/intridea/omniauth/pull/829

I think this PR is still useful for the spec to test the ACS URL fallback behavior, but it may need to wait for a new version of `omniauth` (assuming my PR is accepted).

If the `omniauth` PR is not accepted, I may end up reopening #56 and changing the `deep_dup` to a `dup` or else I'll just move the `initialize_copy` fix into this PR instead.

md5:
https://github.com/intridea/omniauth/pull/829 has been merged! :+1: 

I'll update this PR to bump the `omniauth` requirement once a new version is released and just leave the spec changes for the ACS fallback URL behavior.

md5:
I've updated the `gemspec` to require `omniauth ~> 1.3`. The only code changes in this PR now are for the specs.

jturkel:
:clap:

md5:
I've rebased and force-pushed to this branch to allow it to be merged.

@danwiding @PracticallyGreen/open-source-collaborators Can someone take a look and see if this looks OK?

supernova32:
@md5 I'd say it's a :+1:, specially since the underlying problem was solved directly in the omniauth project. I think we should add a CI solution to this project, though so that we can better test the code.

md5:
@supernova32 I definitely agree about getting the project under CI. It's already got a good start on tests, so it should just be a matter of testing things out under different Ruby interpreters and versions.

I'm not sure what others think in terms of compatibility, but `omniauth` itself is still supporting down to Ruby 1.8.7.

md5:
Looks like `omniauth-oauth2` is also tested in Travis down to Ruby 1.8.7, but I didn't notice a `.travis.yml` in any of the other `omniauth-*` gems maintained by Intridea that have been updated in the last few months.

supernova32:
@md5 let's get this merged and work on adding TravisCI to the project. I'm not sure about supporting 1.8.7 as well, but let's let others chime in as well.

md5:
@supernova32 Thanks for the merge! :+1: 

I've started working on adding Travis over at https://github.com/appropriate/omniauth-saml/tree/travis-ci

md5:
For what it's worth, it looks like `ruby-saml` goes to some effort to support Ruby 1.8.7, so I think `omniauth-saml` probably should as well. I'm in the midst of making that all work in Travis. I expect I'll squash the commits in that branch once I get it working.

supernova32:
@md5 that sounds great! thanks for getting it started.

md5:
No problem.

Right off the bat, I'm seeing some modern hash syntax in [`saml_spec.rb`](https://github.com/PracticallyGreen/omniauth-saml/blob/master/spec/omniauth/strategies/saml_spec.rb), so that would have to be changed if Ruby 1.8.7 is to be supported.

Since Ruby 1.8.7 was EOL'd in July 2014, I'm not sure how important that is to users.

md5:
See #68 for my proposed Travis config :metal:

PR #66, opened 2015-12-23
Update `ruby-saml` to 1.1

supernova32:
Ruby SAML 1.1 was released a couple of months ago and includes some important [changes](https://github.com/onelogin/ruby-saml/blob/master/changelog.md), most importantly the support for SAMLResponse without ds:x509certificate.

This is needed for some environments where GitLab EE is running.

Also bumped the version to 1.4.1 to release a new gem.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,24 +1,24 @@
 PATH
   remote: .
   specs:
-    omniauth-saml (1.3.1)
+    omniauth-saml (1.4.2)
       omniauth (~> 1.1)
-      ruby-saml (~> 1.0.0)
+      ruby-saml (~> 1.1, >= 1.1.1)
 
 GEM
   remote: https://rubygems.org/
   specs:
     diff-lcs (1.2.4)
-    hashie (3.4.2)
+    hashie (3.4.3)
     macaddr (1.7.1)
       systemu (~> 2.6.2)
-    mini_portile (0.6.2)
+    mini_portile2 (2.0.0)
     multi_json (1.3.7)
-    nokogiri (1.6.6.2)
-      mini_portile (~> 0.6.0)
-    omniauth (1.2.2)
+    nokogiri (1.6.7.1)
+      mini_portile2 (~> 2.0.0.rc2)
+    omniauth (1.3.1)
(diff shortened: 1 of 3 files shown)


md5:
I think this should actually be like this:

```ruby
gem.add_runtime_dependency 'ruby-saml', '~> 1.1', '>= 1.1.1'
```

Version 1.1.0 was broken.

supernova32:
@md5 good point. I updated the commit.

supernova32:
and fixed it again, because I read your comment too fast :stuck_out_tongue:

md5:
:+1:

dosire:
@md5 thanks for the feedback, it would be awesome to get this merged.

supernova32:
@bpedro any chance we can get this merged soon?

supernova32:
@brendankemp any chance we can get this reviewed?

brendankemp:
@danwiding is in charge of this project now. He can get you sorted :)

bufferoverflow:
@danwiding Would you be able to merge this and prepare a new release in the near future?

Thanks a lot!
-roger

supernova32:
I guess our only option going forward is to fork this gem, since we have not heard back from the maintainers and the project seems to be abandoned.

danwiding:
@bufferoverflow @supernova32 @md5 @dosire Hey so I guess I've inherited this. I'll be honest my org has moved away from active maintenance of this gem well before I even joined. I believe @bpedro is the  intended maintainer. At any rate, I just granted each of you write privileges to this repo so as to deal with issues like this in the future. However, as a longer term solution I'd like to figure out what org or user account this gem should belong to. If you'd like to take active ownership speak up.

ilikepi:
If no one organization wants to take over maintainership, it might make sense to follow the model adopted for Ryan Bates' CanCan gem.  In that case, the community basically created its own organization to hold [their fork](https://github.com/CanCanCommunity/cancancan).  They also renamed the gem, presumably since they did not have access to the original on rubygems.org, but perhaps that step would not be necessary in this case.

supernova32:
@danwiding I would love to take over the maintainership of this gem. I commented on the new issue that @ilikepi created.

PR #68, opened 2016-02-06
Add Travis configuration

md5:
Travis configuration is based on intridea/omniauth and onelogin/ruby-saml

The `1.8.7` and `jruby-18mode` builds currently fail due to the use of Ruby 1.9+ hash syntax in `saml_spec.rb`. I can correct that if supporting Ruby 1.8.7 is deemed to be a priority, otherwise I can drop 1.8 support.

Also, I originally copied `bundler_args: --without development` from `omniauth`'s Travis config, but there are a number of development dependencies in the `gemspec` file that are actually test dependencies. These would need to all be moved to the `Gemfile` instead to allow `bundler_args: --without development` to work, but I'm not sure there's much value in that.

It may also be desirable for me to drop the `BUNDLED_WITH` changes in the `Gemfile.lock` from this PR.

The test builds can be seen at https://travis-ci.org/appropriate/omniauth-saml

In order to enable Travis for this repo, someone with admin rights will need to make that change. I'd suggest turning on Travis support with the "Build only when .travis.yml is present" flag.


md5:
Note, `ruby-head` is also failing, though I don't have time today to figure out why that's the case.

supernova32:
@md5 I hate the old hash syntax :stuck_out_tongue: Does 1.8.7 really have to be supported? I wonder how many people are still using it. As you mentioned, it reached EOL some 18 months ago.

md5:
@supernova32 If it's only a matter of using the old `Hash` syntax, then it seems like a minor compromise. If it's a significant effort beyond that, then it may indeed be too much to ask of a community-supported gem.

I suspect that as old as it is, Ruby 1.8.7 is still the norm in some of the corporate environments where the `omniauth-saml` gem is most useful.

md5:
I've put a bit of work into trying to make Ruby 1.8.7 work, but there appear to be some non-trivial failures: https://travis-ci.org/appropriate/omniauth-saml/jobs/107807605

For what it's worth, I recall that `omniauth` itself had failing Travis builds on 1.8.7 around the time that I got https://github.com/intridea/omniauth/pull/829 merged. These have since been fixed, but it's not immediately apparent to me which commits fixed them. These are the commits taht happened after my PR there was merged: https://github.com/intridea/omniauth/compare/32c4165919ab34189faa443cea66dfb11ba4fbff...master

However, I think this may have been the main fix (limiting Hashie to an older version on Ruby 1.8): https://github.com/intridea/omniauth/commit/32b5fc815dd2e2dd7cd9a20235c46f7e2cbd85de

I'd have to add some debugging to `.travis.yml` to dump the `Gemfile.lock` for the 1.8 builds to see which Hashie it ends up using, since I'm not able to get a working 1.8.7 build locally under `rbenv`/`ruby-build`.

md5:
In the meantime, it may be prudent to move the 1.8 builds to the allowed failures list and open an issue to investigate Ruby 1.8.7 compatibility.

md5:
I found a bit more time to work on this and I have Ruby 1.8.7 failures down to a single failure after restricting to Hashie `~> 2.0.5`: https://travis-ci.org/appropriate/omniauth-saml/jobs/107817205

The weird thing is that I don't get this failure locally, so I'm not exactly sure what the source of the failure is.

md5:
The issue seems to be a compressed response that it being passed to REXML as if it is an uncompressed response, but it's strange that the test passed for me locally.

supernova32:
@md5 thanks for the work here. I'd say we are good to merge here. All tests, except for the one on 1.8.7 are green. It is really weird that the tests pass locally, but not on travis. 

I'll merge this one so we can continue the discussion about 1.8.7 support.

md5:
Sounds good. I've opened #70 to discuss Ruby 1.8.7 compatibility.

md5:
I've pushed https://github.com/omniauth/omniauth-saml/commit/e86725772f3d5ac11e6576f1935e72552ed99ee2 to allow Ruby 1.8 builds on Travis to be allowed failures so that we can get green tests if all the other builds pass.

supernova32:
:+1:

PR #71, opened 2016-02-09
Add Coveralls integration

md5:


supernova32:
:+1: 

@md5 seems like there is failed build for Ruby 2.3.0 https://travis-ci.org/omniauth/omniauth-saml/jobs/108078525

Can you fix that build? I'll merge it when that one is green.

md5:
@supernova32 I've updated the Travis config to use an `include` matrix instead of an `exclude` matrix. The Ruby 1.8.7 gemfile is now used only by the Ruby 1.8 builds.

See https://travis-ci.org/omniauth/omniauth-saml/builds/108093096

supernova32:
Awesome! :+1:

PR #72, opened 2016-02-09
Support custom attributes

### LINKED ISSUE #45: name of email attribute in SAML callback response isn't expected by saml.rb

md5:
This PR replaces #51. Here is @OsQu's description from that PR:

> Attribute statements might not always have a Name="firstName" attribute, and modifying the idp to support can be an exhausting task (e.g. ADFS).
> 
> Now one can give attribute_statements option, which contains a mapping from statements to an info hash. For example the default mapping is following:

> ```
>   option :attribute_statements, {
    name: ["name"],
    email: ["email", "mail"],
  }
> ```
> This we can support whatever weird configurations the other end might have, without hard-coding them directly to the source.

Closes #45, #49, #51

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -114,7 +114,7 @@ module OmniAuth
           [key, attribute]
         end
 
-        found_attributes.to_h
+        Hash[found_attributes]
       end
 
       extra { { :raw_info => @attributes } }
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -15,6 +15,12 @@ module OmniAuth
         { name: 'last_name', name_format: 'urn:oasis:names:tc:SAML:2.0:attrname-format:basic', friendly_name: 'Family name' }
       ]
       option :attribute_service_name, 'Required attributes'
+      option :attribute_statements, {
+        name: ["name"],
+        email: ["email", "mail"],
+        first_name: ["first_name", "firstname", "firstName"],
+        last_name: ["last_name", "lastname", "lastName"]
+      }
 
       def request_phase
         options[:assertion_consumer_service_url] ||= callback_url
@@ -103,15 +109,23 @@ module OmniAuth
       uid { @name_id }
 
       info do
-        {
-          :name  => @attributes[:name],
-          :email => @attributes[:email] || @attributes[:mail],
-          :first_name => @attributes[:first_name] || @attributes[:firstname] || @attributes[:firstName],
-          :last_name => @attributes[:last_name] || @attributes[:lastname] || @attributes[:lastName]
-        }
+        found_attributes = options.attribute_statements.map do |key, values|
+          attribute = find_attribute_by(values)
+          [key, attribute]
+        end
+
+        found_attributes.to_h
       end
 
       extra { { :raw_info => @attributes } }
+
+      def find_attribute_by(keys)
+        keys.each do |key|
+          return @attributes[key] if @attributes[key]
+        end
+
+        nil
+      end
     end
   end
 end
--- a/README.md
+++ b/README.md
@@ -100,6 +100,11 @@ The service provider metadata used to ease configuration of the SAML SP in the I
 
 * `:attribute_service_name` - Name for the attribute service. Defaults to `Required attributes`.
 
+* `:attribute_statements` - Used to map Attribute Names in a SAMLResponse to
+  entries in the OmniAuth [info hash](https://github.com/intridea/omniauth/wiki/Auth-Hash-Schema#schema-10-and-later).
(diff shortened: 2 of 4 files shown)


bufferoverflow:
:+1:

OsQu:
:+1:!

supernova32:
@md5 Great! thanks for resubmitting. Could we add a hint about this new option to the README as well? I think it should be well documented that it is possible to assign custom attributes and that changes to the idP server are not needed.

md5:
Since this is from @ilikepi's branch, any additions would need to happen there unless we open yet another PR. Perhaps @ilikepi has some time to make those addition or someone can submit a PR to his branch.

ilikepi:
I added a brief entry to the README.  Feedback welcome...I tried to be concise even though this particular option is a bit difficult to understand without some minimal amount of background knowledge.

md5:
@ilikepi Looks good to me! :+1:

lib/omniauth/strategies/saml.rb
```diff
@@ -103,15 +109,23 @@ def other_phase
...
       info do
-        {
-          :name  => @attributes[:name],
-          :email => @attributes[:email] || @attributes[:mail],
-          :first_name => @attributes[:first_name] || @attributes[:firstname] || @attributes[:firstName],
-          :last_name => @attributes[:last_name] || @attributes[:lastname] || @attributes[:lastName]
-        }
+        found_attributes = options.attribute_statements.map do |key, values|
+          attribute = find_attribute_by(values)
+          [key, attribute]
```
md5: Does it make sense to have `[key, nil]` here if no mapped attribute is found? I guess that's how the old code worked, so it's probably fine.

ilikepi:
Updated the entry to avoid repeated use of the word "name".  Ugh...

lib/omniauth/strategies/saml.rb
ilikepi: Hmm...the other option would be to treat it as a validation failure and raise an exception.  That seems overly aggressive, but admittedly I have very little experience with environments in which SAML is used.  There may very well be some environments in which it makes sense to treat a response with a "missing" Attribute as invalid.

md5:
@ilikepi It might be good to give an example with `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress` or another URI-named attribute instead of just a plain name.

lib/omniauth/strategies/saml.rb
md5: I'd still like to see this scheme extended to support multi-valued attributes as suggested in #36, so perhaps a similar approach could be used to add a `:required => true` flag.

ilikepi: Yeah, I wonder if it makes sense to address that in a separate PR.  It's starting to sound like it would be worth considering something more robust than a simple Hash for handling the various mapping scenarios...

md5: I think it should definitely be a separate PR.

supernova32: I agree :smiley:

supernova32:
:+1: I think this is good to merge, since the suggested changes are going to be addressed in a different PR.

PR #78, opened 2016-02-11
feat: create initial CONTRIBUTING.md

### LINKED ISSUE #77: add a CONTRIBUTING.md file
As mentioned within #73 we need a CONTRIBUTING.md file, this has a nice side effect: https://github.com/blog/1184-contributing-guidelines

bufferoverflow:
This closes #77

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,69 @@
+# Contributing
+
+
+## Workflow
+
+We are using the [Feature Branch Workflow (also known as GitHub Flow)](https://guides.github.com/introduction/flow/),
+and prefer delivery as pull requests.
+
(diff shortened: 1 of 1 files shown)


md5:
I'm not a fan of the `type:` annotation on commits. I think it's sets a higher bar for outside contributions than is desirable.

bufferoverflow:
The changelog-generators(e.g. [nodejs variant](https://github.com/ajoslin/conventional-changelog)) are usually configurable, e.g. by default you only have feat and fix to be included within CHANGELOG.md this reduces info on style changes, etc.

Outside contributers are usually familiar with this quickly, they have a look at the commit log and see what type to use and if not we can stash and amend.

supernova32:
If no one has any objections, I'd like to merge this today.

md5:
:+1:

Issue #67, opened 2016-02-02
Future maintainership


ilikepi:
> However, as a longer term solution I'd like to figure out what org or user account this gem should belong to. If you'd like to take active ownership speak up.

The current maintainer is @danwiding, but in [a comment](https://github.com/PracticallyGreen/omniauth-saml/pull/66#issuecomment-178816302) on #66 he stated a desire to discuss a possible transfer of that responsibility.  I figured it makes sense to hold that discussion in its own issue, so here we are.

Here was my initial response, in full:

> If no one organization wants to take over maintainership, it might make sense to follow the model adopted for Ryan Bates' CanCan gem. In that case, the community basically created its own organization to hold [their fork](https://github.com/CanCanCommunity/cancancan). They also renamed the gem, presumably since they did not have access to the original on rubygems.org, but perhaps that step would not be necessary in this case.

supernova32:
@danwiding I would like to take maintainership of this gem. I don't think renaming it's the best way to go, since all omniauth related gems follow the same naming convention.

@danwiding I also checked on RubyGems, and I don't have access there to release a new version. Do you think I could get access there as well?

danwiding:
Hey @supernova32, @brendankemp Just granted you access to RubyGems as well. And thanks for being the first volunteer :) 

We should probably give the other folks a chance to respond for the sake of openness but I'll otherwise start the transfer process by end of week.

supernova32:
@danwiding that is great. I'll cut a new version of the gem first thing tomorrow morning. Let me know if you need anything from me for the move, if nobody else speaks up :smile:

bufferoverflow:
I've just created https://github.com/orgs/omniauth/ so omniauth-saml can be moved later on to a new location.
@supernova32 @brendankemp are added as owners as well.

md5:
@bufferoverflow That seems like an overly broad name unless the folks at Intridea have any interest in sharing that namespace and moving `intridea/omniauth` and the other `intridea/omniauth-*` gems there.

bufferoverflow:
@md5 just added you and @dosire as well.

Yes, it is a broad name. However it's not a problem to have just an omniauth-saml repo there.
The advantage of such a broad name would be that other authentication strategies can join as well and of course omniauth itself as well if they like.

dosire:
I like it.

supernova32:
@bufferoverflow I seem to only have member access to the `omniauth` org. Can you correct that?

[...]

PR #80, opened 2016-02-11
docs: move license from README.md to LICENSE.md

bufferoverflow:
A widely adopted best practice is having a LICENSE.md file, so I've moved the license out from README.md to a dedicated file.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,25 @@
+# License
+
+Copyright © 2016 Omniauth-SAML maintainers
+
+Copyright © 2011-2014 [Practically Green, Inc.](http://www.practicallygreen.com/).
+
+All rights reserved. Released under the MIT license.
+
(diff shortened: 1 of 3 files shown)


md5:
How about adding a link to the license file in `README.md`? Other that that, this looks good to me :+1:

bufferoverflow:
Here is the updated commit, what about the Authors at the end of README.md and within omniauth-saml.gemspec ? Shall we combine this somehow? Any idea?

Shall omniauth-saml.gemspec shall contain the same people as on https://rubygems.org/gems/omniauth-saml ?

md5:
I'm not sure who should be listed as authors in the Gemspec and on Rubygems.org. I don't think that needs to be resolved in this PR. Perhaps it would make sense to discuss that at #67 instead.

bufferoverflow:
@md5 yes, you are right that's a separate topic. So I suggest you merge this.

PR #86, opened 2016-02-27
fix: security issues caused by outdated dependencies

bufferoverflow:
gemnasium reported some [issues](https://gemnasium.com/omniauth/omniauth-saml/alerts)


omniauth-saml.gemspec
```diff
@@ -12,11 +12,12 @@ Gem::Specification.new do |gem|
   gem.homepage      = 'https://github.com/omniauth/omniauth-saml'
 
   gem.add_runtime_dependency 'omniauth', '~> 1.3'
-  gem.add_runtime_dependency 'ruby-saml', '~> 1.1', '>= 1.1.1'
+  gem.add_runtime_dependency 'ruby-saml', '~> 1.1'
```
md5: Can you revert this line? There were some issues with ruby-saml 1.1.0, so I consider this a regression.

bufferoverflow:
updated the PR, coverage decreased because of newer coverage gem.

md5:
@bufferoverflow This looks OK to me. One thing I'm not sure about is whether `coveralls` should be removed from the `Gemfile`. I put it over there so that it could be in the `test` group instead of `development`, but I'm not sure it matters.

bufferoverflow:
The gem specification has only [add_development_dependency](http://guides.rubygems.org/specification-reference/#add_development_dependency) so I guess we should just merge this.

supernova32:
Thanks @bufferoverflow for updating the dependencies. I agree we can merge this.

md5:
@bufferoverflow Sounds good. That being said, I think that having `coveralls` in the `Gemfile` as well as the `*.gemspec` is now redundant.

PR #62, opened 2015-10-27
Allow newer versions of ruby-saml

alex:

PR #89, opened 2016-03-05
fix: security issues caused by outdated dependencies

bufferoverflow:
- remove coveralls from gemspec
- add versions for coveralls and mime-types to Gemfile
- update rack-test to 0.6.3
- remove Gemfile.lock (It seems to be bad practice to checkin the Gemfile.lock when working on a gem.)

this should address the issues mentioned by @md5


bufferoverflow:
@md5 Could you have a look at this?

.gitignore
```diff
@@ -1,3 +1,4 @@
+Gemfile.lock
```
supernova32: @bufferoverflow can you provide references that mention that committing the `.lock` file is bad practice?

bufferoverflow: I talked with some other Ruby coders about this and here is another reference about it:
http://stackoverflow.com/questions/4151495/should-gemfile-lock-be-included-in-gitignore

md5: I've also observed that excluding `Gemfile.lock` from gem repositories is common.

I just did a quick spot check and it looks like `rails` includes the `Gemfile.lock`, while `rack` and `ominiauth` do not.

One advantage of excluding it is that CI builds will always pull the latest dependencies allowed by the ranges, which can result in finding bugs introduced by new versions of dependencies. On the other hand, not having a `Gemfile.lock` means that you could no longer be testing against an older supported version of a dependency, resulting in possible regressions.

I think that to avoid the latter problem, the best practice I've seen is to have special Gemfile with stricter ranges for testing different (usually major) versions of dependencies. The `omniauth` gem, as an example, has separate Gemfiles for testing `rack` 1.3 and `rack`'s `master` branch in addition to its standard `Gemfile`.

md5:
@bufferoverflow What is the version bump for `ruby-saml` about? The reason it has `>= 1.1.1` in there is that version `1.1.0` had a bug that ended up being reported here. See #62, https://github.com/onelogin/ruby-saml/issues/275, and https://github.com/onelogin/ruby-saml/pull/282

If there are no compatibility or security reasons to up the dependency to 1.1.2, I'd drop that change.

Otherwise, this looks good to me.

bufferoverflow:
reason was this: https://gemnasium.com/omniauth/omniauth-saml/alerts
Personally I think updating the dependencies makes always sense as long as they do not break anything.

md5:
@bufferoverflow I don't see any advisories there for `ruby-saml`. As for updating dependencies, I think that *applications* should regularly update their dependencies, but I don't think that infrastructure gems should be forcing their users to upgrade their dependencies unless there's a reason.

bufferoverflow:
@md5 Ok, you convinced me, reworked commit is here.

md5:
:+1:

md5:
@supernova32 I can merge this if you're OK with the `Gemfile.lock` ignoring.

supernova32:
@md5 I liked your suggestion:

> I think that to avoid the latter problem, the best practice I've seen is to have special Gemfile with stricter ranges for testing different (usually major) versions of dependencies. The omniauth gem, as an example, has separate Gemfiles for testing rack 1.3 and rack's master branch in addition to its standard Gemfile.

We could implement more test suites with that, to always test the latest version of dependencies and also to have a locked test that targets the versions we expect to have.

So :+1: for merging.

PR #90, opened 2016-03-05
chore: remove ruby 1.8.7 support

### LINKED ISSUE #70: Determine whether Ruby 1.8.7 should be supported
In #68, we were able to establish that Ruby 1.8.7 should probably work. When I run the specs locally, the Ruby 1.8 builds pass, but they are failing on Travis for some unknown reason.

Given that Ruby 1.8.7 was EOL'd in July 2014, it may not be worth maintaining compatibility. However, both `omniauth` and `ruby-saml` are making an effort to stay compatible with Ruby 1.8.7 and it doesn't seem to involve much of an effort (assuming the tests would pass), so it could be helpful to some users to keep things working for 1.8.7.

bufferoverflow:
This closes #70

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ language: ruby
 script: bundle exec rspec
 matrix:
   include:
-    - rvm: 1.8.7
-      gemfile: gemfiles/ruby-1.8.7.gemfile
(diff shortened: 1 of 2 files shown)


supernova32:
:+1:

PR #91, opened 2016-03-09
chore: remove allow_failures, they pass now always

bufferoverflow:


md5:
I copied this `allow_failures` list from `omniauth`. I was assuming the reason that they were allowing failures for these Ruby versions is that they're more prone to breaking changes that are unrelated to the gem itself. (at least `jruby-head` and `ruby-head`). That being said, I'm fine with removing `allow_failures` here. If these tests become noisy for some reason, it can always be added back.

supernova32:
:+1:

PR #92, opened 2016-03-14
Extract metadata route logic to a separate method

rnubel:
This is a trivial change that will make it easier for [omniauth-multi-provider-saml](https://github.com/salsify/omniauth-multi-provider-saml) to override the method, so it can properly route `/auth/saml/:provider/metadata` to display per-provider metadata.

Some background: for multi-provider SAML, each SAML provider has a unique identifier that scopes all the omniauth paths for it -- e.g. `/auth/saml/myprovider/callback`, `/auth/saml/otherguys/callback`. But the metadata route is broken, as this gem hard-codes the expected path to just `/auth/saml/metadata`. So short of copying the entire `other_phase` method, there's no good workaround (multi-provider-saml passes in a proc value for `request_path`, which `Omniauth::Strategy` [just ignores](https://github.com/intridea/omniauth/blob/32c4165919ab34189faa443cea66dfb11ba4fbff/lib/omniauth/strategy.rb#L380) for literal path comparison).

I think this PR is the easiest way to address the problem, though it'll need to be accompanied by a gem release so that multi-provider-saml can actually depend on it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ env:
     - JRUBY_OPTS="$JRUBY_OPTS --debug"
 language: ruby
 script: bundle exec rspec
+before_install: gem update bundler
 matrix:
   include:
     - rvm: 1.9.3
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -96,8 +96,12 @@ module OmniAuth
         Digest::SHA1.hexdigest(cert.to_der).upcase.scan(/../).join(':')
       end
 
+      def on_metadata_path?
+        on_path?("#{request_path}/metadata")
+      end
+
       def other_phase
-        if on_path?("#{request_path}/metadata")
+        if on_metadata_path?
           # omniauth does not set the strategy on the other_phase
           @env['omniauth.strategy'] ||= self
           setup_phase
--- a/spec/omniauth/strategies/saml_spec.rb
+++ b/spec/omniauth/strategies/saml_spec.rb
@@ -218,4 +218,8 @@ describe OmniAuth::Strategies::SAML, :type => :strategy do
       last_response.body.should match /Required attributes/
     end
   end
+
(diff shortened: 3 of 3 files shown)


rnubel:
No clue why CI is failing just on one jruby build (https://travis-ci.org/omniauth/omniauth-saml/jobs/115955111). I don't feel like it's related to my change.

bufferoverflow:
I guess there is a relation because all previous builds [succeded](https://travis-ci.org/omniauth/omniauth-saml/builds).

One option might be to define fixed jruby versions within `.travis.yml`, see http://rubies.travis-ci.org/

md5:
@bufferoverflow @rnubel Looks like this might be an issue with Bundler in Travis: https://github.com/bundler/bundler/issues/3558#issuecomment-171347979

The weird thing is that the suggestion in that comment is to use the container-based infrastructure with `sudo: false` in the `.travis.yml`, which we're already doing.

I've restarted that Travis build job to see if it passes, since this seems unrelated to the changes in this PR.

md5:
Reading through some of the commits that reference bundler/bundler#3558, it looks like we may want to upgrade bundler in our builds.

I see that @rnubel has just done this in 20650c2488931a09537f99f50042d875d3e25aef

md5:
Looks good to me :+1:

bufferoverflow:
cool, upgrade bundler is the trick.

@rnubel could you update the commit message of 20650c2488931a09537f99f50042d875d3e25aef  to something like this:
```
chore(travis-ci): update bundler before install

Attempt to fix bundler issue caused by bundler/bundler#3558. 
```
So we are inline with the [Conventional Changelog](https://github.com/omniauth/omniauth-saml/blob/master/CONTRIBUTING.md#git-commit-conventions) and I commit this.

rnubel:
@bufferoverflow done! Thanks for helping out, @md5.

md5:
:+1:

PR #95, opened 2016-04-07
Register subclasses with OmniAuth.strategies

md5:
This PR updates `OmniAuth::Strategies::SAML` to register any subclasses as strategies with OmniAuth. This parallels what's done in the `omniauth-oauth2` gem (cf. [here](https://github.com/intridea/omniauth-oauth2/blob/26152673224aca5c3e918bcc83075dbb0659717f/lib/omniauth/strategies/oauth2.rb#L17-L19)).

### DIFF
--- a/spec/omniauth/strategies/saml_spec.rb
+++ b/spec/omniauth/strategies/saml_spec.rb
@@ -222,4 +222,11 @@ describe OmniAuth::Strategies::SAML, :type => :strategy do
   it 'implements #on_metadata_path?' do
     expect(described_class.new(nil)).to respond_to(:on_metadata_path?)
   end
+
+  describe 'subclass behavior' do
+    it 'registers subclasses in OmniAuth.strategies' do
+      subclass = Class.new(described_class)
+      expect(OmniAuth.strategies).to include(described_class, subclass)
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


spec/omniauth/strategies/saml_spec.rb
```diff
@@ -222,4 +222,13 @@ def post_xml(xml=:example_response)
   it 'implements #on_metadata_path?' do
     expect(described_class.new(nil)).to respond_to(:on_metadata_path?)
   end
+
+  describe 'subclasses' do
+    let(:subclass) { Class.new(described_class) }
```
supernova32: instead of declaring the `subclass` in a `let` can we just initialize it within the `it` block?

```ruby
it 'registers subclasses in OmniAuth.strategies' do
  subclass = Class.new(described_class) # or described_class.new
  expect(OmniAuth.strategies).to include(subclass)
end
```

I find this approach easier to read as it defines everything the test needs within the test itself.

md5: Updated and amended. I also threw `described_class` into the `include` check since `omniauth-oauth2` is doing that check and it seems reasonable.

md5:
These `jruby-head` and `ruby-head` Travis failures are annoying...

supernova32:
@md5 I agree. We should move them back to the `allowed_failures`.

md5:
@supernova32 Updated in f74d983 to allow failures of `jruby-head` and `ruby-head`.

PR #99, opened 2016-05-24
Support for sp/idp-initiated single logout fixes #81

### LINKED ISSUE #81: Support single signout
I'm hoping to get single signout working with Omniauth-saml. My understanding is that ruby-saml supports this now, but as far as I can tell, this project does not. Is this something that could be added? (Or documented?)

sanderhahn:
* This pull requests adds the functionality for SLO support (discussed in issue: #81).
* Tested on OneLogin by setting the `idp_slo_target_url` in the configuration (which is not automatically part of the metadata.xml provided by OneLogin).
* <s>Support for downloading the configured certificate, for identity providers that require manual certificate upload.</s>


md5:
@sanderhahn Thanks for your contribution! Do you think you could spend some time enhancing the specs for this important new feature?

lib/omniauth/strategies/saml.rb
```diff
@@ -67,21 +63,67 @@ def callback_phase
...
+                logout_request_id = logout_request.id
+                logout_response = OneLogin::RubySaml::SloLogoutresponse.new.create(settings, logout_request_id, nil, RelayState: request.params["RelayState"])
+                return redirect(logout_response)
+              end
+            end
+          end
+          raise OmniAuth::Strategies::SAML::ValidationError.new("IdP initiated LogoutRequest was not valid!")
+        end
+
+        unless request.params['SAMLResponse']
```
supernova32: Instead of using an unless here and checking for `request.params["SAMLRequest"]` again, could you use an else clause from the original `if` and raise the exception there?

md5:
As a first pass review, I think it would be better to separate the logout handling into a separate route instead of adding conditional logic to `/callback`. Perhaps `/slo` or `/logout`?

md5:
I also don't fully understand the use case for `/certificate`. Would you mind elaborating why you think that's necessary? It seems like someone who configured the certificate in the first place would already have access to the certificate without needing to download it from the application. It's also included in the `/metadata` document already.

sanderhahn:
- Yes, i can see if i can add some specs if this helps getting the functionality into `omniauth-saml`.
- Can move the code to an "slo" endpoint if you want (this will also reduce the mixup of SSO/SLO logic in the `callback_phase`): then i can use the `on_path?` dispatching in `other_phase`. However in the `metadata.xml` i only see the `AssertionConsumerService` tag that contains the request url. Where would the slo url fit into the metadata and does this require an update of `ruby-saml` settings?
- Added the certificate endpoint because we use `omniauth-saml` in a multitenant setting (using `config.omniauth :saml, setup: SETUP_OMNIAUTH_SAML_PROC`. The certificate is automatically generated for organizations that want to use SAML. Understand this use-case might be different and i can remove the endpoint if you want. Some SAML providers seems to want you to upload the certificate in their backend and this is just a convenience method.

sanderhahn:
In the `sessions_controller` in our application i use this code in the `destroy` action. Feels that it should be part of `omniauth-saml` but i am unsure if there is a place within `omniauth` where it should hook into or that it should go in some util method:

```ruby
  def sp_logout_request(settings, saml_uid)
    # LogoutRequest accepts plain browser requests w/o paramters
    if ! settings.idp_slo_target_url.nil?

      # Since we created a new SAML request, save the transaction_id
      # to compare it with the response we get back
      logout_request = OneLogin::RubySaml::Logoutrequest.new()

      saml_transaction_id = logout_request.uuid
      logger.info "New SP SLO for userid '#{saml_uid}' transactionid '#{saml_transaction_id}'"

      if settings.name_identifier_value.nil?
        settings.name_identifier_value = saml_uid
      end

      redirect_to(logout_request.create(settings, RelayState: root_url))
      true
    else
      false
    end
  end
```

sanderhahn:
Tried to improve the code coverage of the specs by adding some SLO examples. Also introduced an `slo` end-point to initiate the service provider SLO. One extra option is `relay_state` as a default, because sometimes the IdP doesn't use provide the RelayState value.

bufferoverflow:
Great work!

Would be awesome if you could use the commit conventions mentioned here: https://github.com/omniauth/omniauth-saml/blob/master/CONTRIBUTING.md#git-commit-conventions

sanderhahn:
Okay i have squashed the commits and applied your commit convention :)

md5:
@sanderhahn It looks like you're pulling in amended versions of a couple older commits. I think you need to rebase against a current copy of `master` from this repo.

lib/omniauth/strategies/saml.rb
```diff
@@ -100,25 +95,41 @@ def response_fingerprint
         Digest::SHA1.hexdigest(cert.to_der).upcase.scan(/../).join(':')
       end
 
-      def on_metadata_path?
```
md5: Removing `on_metadata_path?` is a breaking change for anyone who has customized that method in a subclass.

lib/omniauth/strategies/saml.rb
```diff
@@ -27,6 +27,7 @@ def self.inherited(subclass)
         first_name: ["first_name", "firstname", "firstName"],
         last_name: ["last_name", "lastname", "lastName"]
       }
+      option :relay_state
```
md5: I think `:default_relay_state` could be a better name. It may make sense to allow it to be a proc that takes the request as a parameter as well, though I'm not sure exactly how it would be used.

md5: Also, adding it as an option like this will mean that it gets passed to `ruby-saml` as a setting. It may make sense to `delete` it from the options before construction the `Settings` object.

md5:
Regarding getting the SLO path into the metadata, that appears to be done here: https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/metadata.rb#L56-L62

It's based on setting `single_logout_service_url` and `single_logout_service_binding`.

Also, looking at that metadata code, it seems like the `SAMLRequest` and `LogoutResponse` cases are expected to be handled by the same URL. In your current code, you're processing `LogoutResponse` at the ACS URL. I'd have to look at the specs to see if this is expected behavior for an SP.

md5:
Also, I do think the `/certificate` route should be removed since that doesn't seem like a common use case.

sanderhahn:
Thanks, i'll see if i can get these things fixed tomorrow.

md5:
If others feel strongly that the `/certificate` route is useful, I'm willing to cave on that, but I don't think it belongs in this PR regardless since it is a separate feature.

lib/omniauth/strategies/saml.rb
sanderhahn: Options that are not for `ruby-saml` settings are also passed along to the `Settings` object, but the initializer seems to ignore any that don't have a setter. Could filter the options using `OneLogin::RubySaml::Settings.instance_methods.select { |name| name.to_s.end_with?("=") }` but this seems kindof redundant.

bufferoverflow:
I guess the only missing thing is adding `idp_slo_target_url` to the `README.md` and ready to merge.

@md5 what do you think?

md5:
@bufferoverflow I haven't had a chance to actually test this out and I'm not sure I can do so easily with our corporate IdP (I'd have to check with the team that runs it whether they're allowing SLO configuration).

In addition to documenting `idp_slo_target_url`, the `default_relay_state` option also needs to be documented.

Another thing I think we need to look closely at is the use of `session`, since this strategy didn't previously do anything with the session directly (although `omniauth` itself does).

If we could get another set of eyes on this for a technical review besides me, I think that would be wise.

sanderhahn:
Have tested the SLO interactively using a OneLogin free trail. The SP initiated flows also seemed to work on Orka, ADFS and Azure AD.

bufferoverflow:
@sanderhahn thank you very much for updating the docs and providing the test results

I'm unable to test on our environment at the moment. I think we need another voice before we merge.

grekko:
@sanderhahn Thanks for your effort. I am currently using your omniauth-saml branch in a project and am looking forward to have it merged into master.

md5:
Sorry I haven't been able to test these changes. It turns out our corporate IdP does not configure single log out as a matter of policy.

md5:
@supernova32 Any chance you can give this another code review in addition to what I've done?

README.md
```diff
@@ -143,6 +150,32 @@ end
...
+```ruby
+class SessionsController < Devise::SessionsController
+  # ...
+
+  def destroy
+    # Preserve the saml_uid in the session
+    saml_uid = session["saml_uid"]
+    super do
+      session["saml_uid"] = saml_uid
+      if ! SAML_SETTINGS.idp_slo_target_url.nil?
```
supernova32: This looks like a double negative and is bound to cause confusion. You are doing `if not nil` which is the same as `if present?`, which is easier to understand. I know it's just the example in the README, but ti should be easy to read.

Please change it to:

```ruby
if SAML_SETTINGS.idp_slo_target_url.present?
```

pitbulk: :+1: +1

sanderhahn: Actually `present?` is a Rails helper and this project does not directly depend on Rails. Shall i change it in `unless`?

bufferoverflow: I would say we keep it as it is an merge. 
ok? @md5 @supernova32

md5: @bufferoverflow I think that `unless` is better than `if !`, but I don't consider it a blocker. That being said, I'm not comfortable merging this without a more affirmative sign-off from @supernova32, given that I haven't been able to test this feature due to our corporate IdP's lack of support for SLO.

bufferoverflow: Merge and release, we can fix fast and publish a new minor release if required.

I would love to test this on our next GitLab deplyoment!

supernova32: Let's merge, release, and test, then. I'll prepare the release. Should it be `1.7.0`?

bufferoverflow: yes, 1.7.0

supernova32:
@sanderhahn can you resolve the conflicts?

sanderhahn:
Sure no problem! Will replace the `if/nil?`/`unless` and will try to get conflicts resolved. Probably be tomorrow evening or the next evening (CET)...

supernova32:
@sanderhahn sure, no problem 😃

supernova32:
@sanderhahn please take a look at the conflicts when you have time.

drewvs:
sanderhahn, are you still actively working on this?

supernova32:
Replaced by #115

Issue #83, opened 2016-02-19
Add support for parsing remote IdP metadata


tomilaine:
Hi,

Is anyone working on adding the metadata exchange feature documented in:
https://github.com/onelogin/ruby-saml#metadata-based-configuration

Being able to parse the remote IdP metadata would help a lot in taking this library into use. Also the SP would be able to react to possible sudden changes in IdP metadata.

Tomi

bufferoverflow:
I was not aware of the remote metadata feature, it would make setup much easier. I'm not aware of someone working on this, feel free to provide a Patch, see [CONTRIBUTING.md](https://github.com/omniauth/omniauth-saml/blob/master/CONTRIBUTING.md).

sanderhahn:
At the moment i am parsing metadata and then extracting the `Settings` properties back into `env["omniauth.strategy"].options`. Configuration would be nicer if its possible to directly assign a `settings` property with `OneLogin::RubySaml::Settings` or a callback for retrieving the settings.

rhoerbe:
A nice feature of the configuration-by-metadata is the option to do local idp-discovery, particularly if the IDP metadata has mdui_extensions provided.

peterwake:
If anyone is looking to parse Microsoft federationmetadata.xml, I have written a gist that is quite rough and ready but gets you the key data you will need.. https://gist.github.com/peterwake/18349642f8f16397d1812461bac11aff

PR #100, opened 2016-05-31
refactor: allow metadata_xml for settings ruby-saml

sanderhahn:
Pull request intended for discussion/feedback (relates to issue #83). Tests fail at the moment because of https://github.com/onelogin/ruby-saml/pull/331.

Adds an metadata_xml that can be used for feeding the metadata xml of the identity provider. Refactors the settings to have a single point of initialization.


supernova32:
After #115 is merged, this would no longer make sense as is. @sanderhahn if you feel like you still need the functionality you are adding here, please wait for #115 to be merged, and then add the feature missing.

PR #101, opened 2016-06-01
feat: Add the soft option to SAML options

quamen:
When setting up the providr you can now set the soft option.

  provider :saml,
    issuer: 'Example',
    idp_sso_target_url: 'example.com/sso',
    idp_cert: 'mycert',
    soft: true

The default is still false, but you can now over-ride this to turn
off raising of SAML validation errors.

This is useful when testing or in environment where you don't really
want to set up a real SAML server with signed requests etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,9 @@ The service provider metadata used to ease configuration of the SAML SP in the I
 
 * `:uid_attribute` - Attribute that uniquely identifies the user. If unset, the name identifier returned by the IdP is used.
 
+* `:soft` - Used to enable/disable the soft mode. When is set `false` (the
+  default), the SAML validation errors will raise an exception. Setting this to `true` is useful when testing or in environment where you don't want to set up a real SAML server with signed requests. **You should never do this in a production environment**
+
 * See the `OneLogin::RubySaml::Settings` class in the [Ruby SAML gem](https://github.com/onelogin/ruby-saml) for additional supported options.
 
 ## Devise Integration
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -10,7 +10,7 @@ module OmniAuth
         OmniAuth::Strategy.included(subclass)
       end
 
-      OTHER_REQUEST_OPTIONS = [:skip_conditions, :allowed_clock_drift, :matches_request_id, :skip_subject_confirmation].freeze
+      OTHER_REQUEST_OPTIONS = [:skip_conditions, :allowed_clock_drift, :matches_request_id, :skip_subject_confirmation, :soft].freeze
 
       option :name_identifier_format, nil
       option :idp_sso_target_url_runtime_params, {}
@@ -177,7 +177,7 @@ module OmniAuth
       def handle_response(raw_response, opts, settings)
         response = OneLogin::RubySaml::Response.new(raw_response, opts.merge(settings: settings))
         response.attributes["fingerprint"] = options.idp_cert_fingerprint
-        response.soft = false
+        response.soft = opts[:soft] || false
 
         response.is_valid?
         @name_id = response.name_id
--- a/spec/omniauth/strategies/saml_spec.rb
+++ b/spec/omniauth/strategies/saml_spec.rb
@@ -186,6 +186,37 @@ describe OmniAuth::Strategies::SAML, :type => :strategy do
       it { should fail_with(:invalid_ticket) }
     end
 
+    context "when run with soft option" do
+      before :each do
+        saml_options[:soft] = true
+      end
+
+      context "when the fingerprint is invalid" do
+        before :each do
+          saml_options[:idp_cert_fingerprint] = "00:00:00:00:00:0C:6C:A9:41:0F:6E:83:F6:D1:52:25:45:58:89:FB"
+          post_xml
+        end
+
+        it { should_not fail_with(:invalid_ticket) }
+      end
+
+      context "when the digest is invalid" do
+        before :each do
+          post_xml :digest_mismatch
+        end
+
+        it { should_not fail_with(:invalid_ticket) }
+      end
+
+      context "when the signature is invalid" do
+        before :each do
+          post_xml :invalid_signature
+        end
+
+        it { should_not fail_with(:invalid_ticket) }
+      end
+    end
+
     context "when response has custom attributes" do
       before :each do
         saml_options[:idp_cert_fingerprint] = "3B:82:F1:F5:54:FC:A8:FF:12:B8:4B:B8:16:61:1D:E4:8E:9B:E2:3C"


md5:
This looks fine to me. It shouldn't affect existing users. I do think that the option should be documented in `README.md` though.

quamen:
@md5 I've update the README to include the soft option. Let me know if you think it needs more work or if I've missed a place where it should be documented.

README.md
```diff
@@ -125,6 +127,9 @@ The service provider metadata used to ease configuration of the SAML SP in the I
   *Note*: All attributes can also be found in an array under `auth_hash[:extra][:raw_info]`,
   so this setting should only be used to map attributes that are part of the OmniAuth info hash schema.
 
+* `:soft` - Used to enable/disable the soft mode. When is set false (the
```
md5: Can you please put backticks around "false"? Same with "true" below.

README.md
```diff
@@ -125,6 +127,9 @@ The service provider metadata used to ease configuration of the SAML SP in the I
...
+  default), the saml validations errors will raise an exception. Setting this to true is useful when testing or in environment where you don't want to set up a real SAML server with signed requests.
```
md5: Can you change "the saml validations errors will" to "SAML validation errors will"?

README.md
```diff
@@ -39,7 +39,8 @@ use OmniAuth::Strategies::SAML,
   :idp_cert                           => "-----BEGIN CERTIFICATE-----\n...-----END CERTIFICATE-----",
   :idp_cert_fingerprint               => "E7:91:B2:E1:...",
   :idp_cert_fingerprint_validator     => lambda { |fingerprint| fingerprint },
-  :name_identifier_format             => "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress"
+  :name_identifier_format             => "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress",
+  :soft                               => false
```
md5: I don't think we should mention `soft` in these examples.

md5:
Thanks @quamen. I made some minor comments on the copy in the `README.md`. I think those additions are good, but I almost feel that it needs to be more strongly worded to make it *very* clear that `soft` should not be used in a production situation.

Also, would you mind amending your commit to have `chore:` in the commit message? For that matter, it might be good to just squash it into the previous commit.

quamen:
@md5 I've updated the README and added `chore:` to the beginning of the commit message.

There wasn't an existing style for security style warnings that I could find so I went with **bold warning about security** in the README documentation for the soft flag.

lib/omniauth/strategies/saml.rb
```diff
@@ -10,7 +10,7 @@ def self.inherited(subclass)
         OmniAuth::Strategy.included(subclass)
       end
 
-      OTHER_REQUEST_OPTIONS = [:skip_conditions, :allowed_clock_drift, :matches_request_id, :skip_subject_confirmation].freeze
+      OTHER_REQUEST_OPTIONS = [:skip_conditions, :allowed_clock_drift, :matches_request_id, :skip_subject_confirmation, :soft].freeze
```
md5: Is this line necessary? Looking at the `ruby-saml` source, it seems that `soft` gets picked up from `settings`.

md5:
I'm sorry to go back and forth on this, but the more I think about it, the more I think this shouldn't even be mentioned in the `README.md`.

We already support passing through a bunch of properties to `ruby-saml` that are not documented in the `README.md`, so the only thing that's really needed to support allowing `soft` to be used is changing the line that unilaterally sets `response.soft = false` to allow it to be `true`.

md5:
@bufferoverflow @supernova32 Do you guys have an opinion on this?

quamen:
@md5 I'm happy to change the implementation if that is what you and the other maintainers feel is best. Let me know and I'll re-work this pull request.

md5:
@quamen I think the implementation is fine.

I'm assuming the way you came to make this PR was like this:

1. You noticed that `ruby-saml` supports a `soft` option
2. You tried setting `soft: true` in the `omniauth-saml` settings
3. You noticed that it didn't work
4. You looked at the `omniauth-saml` source code and noticed that `soft` is hard-coded to `false`

Assuming that's the case, I think it's fine for other developers to discover that `soft` now works in the same way (once this PR merges).

The only thing I think needs to be done with this is removing `:soft` from `OTHER_REQUEST_OPTIONS`, unless I'm missing something. I'd also like to get agreement from the other maintainers that leaving this option documented only in `ruby-saml` is fine.

quamen:
@md5 that's exactly how I came to make this PR. 

I'll wait to hear from you and the other maintainers, then I'll update the code and/or README to suit.

bufferoverflow:
Im ok with this, the statement within README.md is very clear.

md5:
@bufferoverflow Sounds good

@quamen I think the only unresolved issue in that case is my comment about `OTHER_REQUEST_OPTIONS`. Can you explain why you think that change is necessary or just remove it by amending your commit?

lib/omniauth/strategies/saml.rb
supernova32: Agreed, this should not be here. It's not necessary, since the `soft` option is already passed to `ruby-saml` in the `settings` hash.

quamen:
> @quamen this looks good. Can you please address the comments we left?

@supernova32 I've removed the `:soft` option from the `OTHER_REQUEST_OPTIONS` array.

md5:
@quamen Sorry for the delay on following up on your changes. Do you mind rebasing this PR?

quamen:
> @quamen Sorry for the delay on following up on your changes. Do you mind rebasing this PR?

@md5 I've rebased against master.

md5:
@quamen Looks like there are some spec failures now. Mind taking a look at those?

lib/omniauth/strategies/saml.rb
quamen: This is now required after rebasing master.

quamen:
> @quamen Looks like there are some spec failures now. Mind taking a look at those?

@md5 sorry about that, rebased it all very quickly before getting distracted by other things.

This should go green now.

lib/omniauth/strategies/saml.rb
md5: I just made the following changes to your branch and the specs still passed:

```diff
diff --git a/lib/omniauth/strategies/saml.rb b/lib/omniauth/strategies/saml.rb
index 01f0a59..a82e3c4 100644
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -10,7 +10,7 @@ module OmniAuth
         OmniAuth::Strategy.included(subclass)
       end
 
-      OTHER_REQUEST_OPTIONS = [:skip_conditions, :allowed_clock_drift, :matches_request_id, :skip_subject_confirmation, :soft].freeze
+      OTHER_REQUEST_OPTIONS = [:skip_conditions, :allowed_clock_drift, :matches_request_id, :skip_subject_confirmation].freeze
 
       option :name_identifier_format, nil
       option :idp_sso_target_url_runtime_params, {}
@@ -29,6 +29,8 @@ module OmniAuth
       }
       option :slo_default_relay_state
       option :uid_attribute
+      # NOTE: This setting should *never* be set to true in a production environment
+      option :soft, false
 
       def request_phase
         options[:assertion_consumer_service_url] ||= callback_url
@@ -177,7 +179,6 @@ module OmniAuth
       def handle_response(raw_response, opts, settings)
         response = OneLogin::RubySaml::Response.new(raw_response, opts.merge(settings: settings))
         response.attributes["fingerprint"] = options.idp_cert_fingerprint
-        response.soft = opts[:soft] || false
 
         response.is_valid?
         @name_id = response.name_id
@@ -214,7 +215,6 @@ module OmniAuth
         # the LogoutResponse, verify it, then actually delete our session.
 
         logout_response = OneLogin::RubySaml::Logoutresponse.new(raw_response, settings, :matches_request_id => session["saml_transaction_id"])
-        logout_response.soft = false
         logout_response.validate
 
         session.delete("saml_uid")
```

md5:
@quamen Could you describe the scenario where you want to have `omniauth-saml` in your middleware stack but you don't have an actual SAML server? After looking at this PR again, the big caveat about not using this in production is making me question the need for this change.

In a testing scenario, it seems like it should be possible to generate a usable SAML response with `ruby-saml` and a key pair generated just for purpose of testing or to mock out the response appropriately.

keysen:
Hey, I bump this PR as I'm interested by this option too. Why it's not merged yet?

My usecase is:
We have a staging environment exactly like the production, It's a SaaS product, each product have extensions whose Saml and every time we have a non-technical person who wants to setup SAML first on staging, it's hardcore and we add new SAML extensions quite often. It could help them to have a soft option to be sure everything works before fixing all the errors. Could we have this one?

PR #107, opened 2016-06-27
feat: added the response object to the extra['response_object']

jiongye:
feat: added the Ruby-saml response object to the extra attribute, so we can use it in the controller via "env['omniauth.auth'].extra.response_object"

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -76,6 +76,7 @@ module OmniAuth
 
         @name_id = response.name_id
         @attributes = response.attributes
+        @response_object = response
 
         if @name_id.nil? || @name_id.empty?
           raise OmniAuth::Strategies::SAML::ValidationError.new("SAML response missing 'name_id'")
@@ -135,7 +136,7 @@ module OmniAuth
         Hash[found_attributes]
       end
 
-      extra { { :raw_info => @attributes } }
+      extra { { :raw_info => @attributes, :response_object =>  @response_object } }
 
       def find_attribute_by(keys)
         keys.each do |key|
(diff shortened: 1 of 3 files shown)


md5:
This looks like a good idea to me. I [suggested](https://github.com/omniauth/omniauth-saml/pull/29#issuecomment-182995617) doing this in #29.

lib/omniauth/strategies/saml.rb
```diff
@@ -135,7 +136,7 @@ def other_phase
         Hash[found_attributes]
       end
 
-      extra { { :raw_info => @attributes } }
+      extra { { :raw_info => @attributes, :response_object =>  @response_object} }
```
md5: Nitpick: can you add the missing space before the closing brace (after `@response_object`)

spec/omniauth/strategies/saml_spec.rb
```diff
@@ -114,6 +114,10 @@ def post_xml(xml=:example_response)
           'fingerprint'  => saml_options[:idp_cert_fingerprint]
         }
       end
+
+      it "should set the response_object to the response object from ruby_saml response" do
+        auth_hash['extra']['response_object'].class.should == OneLogin::RubySaml::Response
```
md5: I'd suggest using `.should be_kind_of(OneLogin::RubySaml::Response)`

md5:
@jiongye If you don't mind addressing my comments, that would be great. Please squash your commits and prefix the commit message with `feat:`.

jiongye:
@md5 I made the changes as you suggested. Please let me know if it looks good.

md5:
Thanks @jiongye. This looks good to me.

Any chance you could update the `README.md` as well?

jiongye:
@md5 I updated the README.md, let me know if that looks good to you.

md5:
Thanks @jiongye! 👍

PR #108, opened 2016-06-28
fix: saved validation error to session so we can reuse after redirect

jiongye:
When there is a validation error on the SAML response, Omniauth will redirect the request to the failure endpoint. In the case of redirection, everything saved in the "env" variable is lost because it only exist in the current request. So in order to access the error message in the failure endpoint, we need to use session to save the error message.

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -82,10 +82,12 @@ module OmniAuth
         end
 
         super
-      rescue OmniAuth::Strategies::SAML::ValidationError
-        fail!(:invalid_ticket, $!)
-      rescue OneLogin::RubySaml::ValidationError
-        fail!(:invalid_ticket, $!)
+      rescue OmniAuth::Strategies::SAML::ValidationError => exception
+        session['saml_error'] = exception.to_s
+        fail!(:invalid_ticket, exception)
+      rescue OneLogin::RubySaml::ValidationError => exception
+        session['saml_error'] = exception.to_s
+        fail!(:invalid_ticket, exception)
       end
 
       # Obtain an idp certificate fingerprint from the response.
--- a/spec/omniauth/strategies/saml_spec.rb
+++ b/spec/omniauth/strategies/saml_spec.rb
@@ -161,6 +161,12 @@ describe OmniAuth::Strategies::SAML, :type => :strategy do
       end
 
       it { should fail_with(:invalid_ticket) }
+
+      it "should set session['saml_error'] but not env['omniauth.error'] after redirect" do
+        follow_redirect!
+        last_request.env['omniauth.error'].should be_nil
+        last_request.env["rack.session"]["saml_error"].should == "Fingerprint mismatch"
+      end
     end
 
     context "when the digest is invalid" do
@@ -169,6 +175,12 @@ describe OmniAuth::Strategies::SAML, :type => :strategy do
       end
 
       it { should fail_with(:invalid_ticket) }
+
+      it "should set session['saml_error'] but not env['omniauth.error'] after redirect" do
+        follow_redirect!
+        last_request.env['omniauth.error'].should be_nil
+        last_request.env["rack.session"]["saml_error"].should == "Digest mismatch"
+      end
     end
 
     context "when the signature is invalid" do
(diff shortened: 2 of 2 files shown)


md5:
@jiongye What's the use case for having this error message available after the redirect? These message are generally not something that should be shown to the user.

I feel like this is a specialized use case that calls for as custom `on_failure` handler. In that case, you can get the underlying error from `env['omniauth.error']`: https://github.com/intridea/omniauth/blob/1cc1cf4b2821a7d2a4a376a5ca93c61b6bd8b5f1/lib/omniauth/strategy.rb#L469

jiongye:
@md5 There is a bug when handling the failure case. I know Omniauth save the exception in the env['omniauth.error'], but the env's life cycle is within the request, once you make a redirection, the env is reset to a new one. you can look at the rspec test I added for detail.  The way Omniauth handle the failure is to redirect to a failure endpoint: https://github.com/intridea/omniauth/blob/1cc1cf4b2821a7d2a4a376a5ca93c61b6bd8b5f1/lib/omniauth/failure_endpoint.rb#L19

So that is why the error message get lost after the validation, and I have to use a session to store that information.  

Or maybe I should create a pull request for the Omniauth gem, but that gem has not been active for awhile now.

jiongye:
@md5 Since Omniauth redirect the user to the failure endpoint on validation error, I think it is good to return the error message in the redirected request, so we can do something about it if we want. I think that is the reason they want to save the exception in env['omniauth.error'].  To my own use case, I will return the error message to the user, so they know why their authentication fails, and if they need to contact us for support, they can provide that message to us, so we can trouble shoot it faster.

md5:
@jiongye I wouldn't call that a bug, since the error message is logged internally, but I'm glad we have the same understanding of what's going on here. I also don't think that returning the SAML validation error message to the end user is a typical use case, though I could be wrong.

It seems like it would be very straightforward in your case to either subclass or wrap `OmniAuth::FailureEndpoint` to provide a `call` method that saves the error into the session there.

jiongye:
@md5 I see what you mean with custom `on_failure` callback, if the custom callback is passed in as a block, then the `env['omniauth.error']` will still hold the exception. 

I am not sure if it is a good idea to either subclass or wrap the `OmniAuth::FailureEndpoint` to do that, because if I do that, I will have to either change the Omniauth config class: https://github.com/intridea/omniauth/blob/master/lib/omniauth.rb#L36 to point to the new class, or have to provide the new class in the config setting when initialize the saml provider. I want something that users can use the default setting and everything works fine. 

Maybe I misunderstood what you mean, can you give me more detail on this? Or what is your suggestion for the best solution here?

md5:
@jiongye My suggestion was to set `config.on_failure` to point to the new class.

I guess I had assumed that your users were end users of a particular site with configurable SAML IdPs, but if you're talking about developer end users, it sounds like you might need to have your own gem that extends `omniauth-saml` with some default configuration choices.

jiongye:
@md5 Thanks! I see what you mean. This is really a bug in the Omniauth gem, fixing it here doesn't make sense.

md5:
@jiongye Cool.

I still don't consider this a bug and I think that using the session to pass along arbitrary exception messages for all strategies could be problematic.

jiongye:
@md5 Can you give me an example why you think using session to pass along arbitrary exception messages could be problematic?  Is this a generic case, or is just for the Omniauth gem? I am trying to understand more, so I can have a second thought if I consider to use session to do similar thing in the future. Thanks!

md5:
Session storage can be limited depending on what mechanism is being used (e.g. cookie-based storage), so I think that having something as broadly used as OmniAuth start storing the exception message could end up being an unexpected breaking change for some downstream applications. Also, since changing this at the OmniAuth level would impact all strategies, it's hard to know how big to expect the string generated by `error.to_s` to be for an arbitrary `Error` instance.

I think the the case of `omniauth-saml` itself, the expected error message won't be too large, but it will still result in an increase in required session storage that could push some applications over the edge. Given that I don't think these messages should be shown to users in most cases, I'm opposed to including them in the session from `omniauth-saml` without an explicit developer opt-in.

jiongye:
@md5 Thank you for the explain this in detail, that makes total sense.

jiongye:
This is for documentation on how I am able to fix the issue without changing any code in the gem.
```
# Instead of changing the gem to set the error inside the gem, set the error in the config
OmniAuth.config.on_failure = Proc.new { |env|
  env["rack.session"]['saml_error'] = env['omniauth.error'].to_s
  OmniAuth::FailureEndpoint.new(env).redirect_to_failure
}
```

md5:
Thanks @jiongye. I'm sure this will help someone else that has a similar need in the future.

I'd actually recommend calling `OmniAuth::FailureEndpoint.call(env)` instead to retain the `raise` in development env functionality and any future changes to the default endpoint.

jiongye:
@md5 Yeah, good point. Thanks!

PR #110, opened 2016-09-13
chore: Add conventional changelog generator

toupeira:
This is a spike for @bufferoverflow to see how we can integrate a [conventional changelog generator](https://github.com/conventional-changelog/conventional-changelog) into Ruby projects. Unfortunately there aren't many options, and I already ran into several issues with the [`conventional-changelog`](https://github.com/dcrec1/conventional-changelog-ruby) gem I'm using here:

- it's very basic, e.g. it only detects `feat` and `fix`
- the behaviour and CLI arguments are not very intuitive,  it's very easy to end up with commits listed under the wrong version header, and it's not possible to generate entries for previous releases either
- there's barely any error handling, wrong CLI arguments can result in an empty `CHANGELOG.md` file

I looked around a bit more for alternatives but couldn't find any other conventional changelog generators written in Ruby, only some older gems for "non-conventional" changelogs (e.g. [vclog](http://rubyworks.github.io/vclog/)). I also stumbled over [clog](http://blog.thoughtram.io/announcements/tools/2014/09/18/announcing-clog-a-conventional-changelog-generator-for-the-rest-of-us.html) which is written in Rust and is a full-featured port of the original NodeJS generator, but they only have Mac binaries so far and I guess a compiled language is overkill for this anyway :) So we'll probably look into improving the gem first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ coverage/
 spec/support/example_private_key.pem
 /gemfiles/*.lock
 .idea/
+.bundle/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,10 @@
-# OmniAuth SAML Version History
-
-A generic SAML strategy for OmniAuth.
-
-https://github.com/omniauth/omniauth-saml
-
-## 1.6.0 (2016-06-27)
+<a name="v1.6.0"></a>
+### v1.6.0 (2016-06-27)
 * Ensure that subclasses of `OmniAuth::Stategies::SAML` are registered with OmniAuth as strategies (https://github.com/omniauth/omniauth-saml/pull/95)
 * Update ruby-saml to 1.3 to address [CVE-2016-5697](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-5697) (Signature wrapping attacks)
 
-## 1.5.0 (2016-02-25)
+<a name="v1.5.0"></a>
+### v1.5.0 (2016-02-25)
 
 * Initialize OneLogin::RubySaml::Response instance with settings
 * Adding "settings" to Response Class at initialization to handle signing verification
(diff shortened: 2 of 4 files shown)


bufferoverflow:
Great work, I think that's the direction we should take.

Could you please figure out what happend with the builds ran by travis?

toupeira:
@bufferoverflow it was failing on older Ruby versions, I updated the dependencies now to support them again.

md5:
@toupeira Those dependency changes are not going to be merged. We need to support the newer versions as well as the older versions.

@supernova32 and I were having a discussion around getting Travis working again over in #111

toupeira:
@md5 ok thanks for the heads up, I removed that commit from the PR.

md5:
@toupeira Thanks. We may need to rebase this once I get the Travis failures sorted out.

md5:
See https://github.com/omniauth/omniauth-saml/pull/113 for the `appraisal` changes to fix Travis.

md5:
@toupeira If you rebase this against `master`, the tests should now pass.

md5:
I did try to restart the Travis build, since I thought that would result in it testing a merge against the current `master`, but I guess `+refs/pull/110/merge` doesn't mean what I thought it means.

toupeira:
@md5 ok, rebased and tests are passing again.

bufferoverflow:
:+1: let's merge this

md5:
Why is CHANGELOG.md showing v1.7.0 with a date in May 2016 when that tag hasn't been made yet? Shouldn't it say “unreleased” or something?

toupeira:
@md5 ah you're right that's not supposed to be there, I only wanted to update the formatting in `CHANGELOG.md` so it matches the generated one. I removed v1.7.0 again, so the idea is to wait with updating the changelog until the next release is ready.

toupeira:
@md5 I only removed the `v1.7.0` section, the initial text was already gone because the gem always prepends to the beginning of the file, and similarly I changed the formatting to match the gem's output.

I'm not happy with this either, and we've actually decided to write our own gem now to avoid such issues. I'll leave it up to you and @bufferoverflow if you want to merge this PR as is, or close it and wait for a better one :)

PR #113, opened 2016-09-16
Fix Travis builds using appraisal gem

md5:
This PR fixes the Travis builds by using different gemfiles for different versions of Ruby and some gem dependencies (notably `rack` 1.x and 2.x). These gemfiles are managed by the `appraisal` gem.

It also switches our build matrix from testing with `jruby-head` to testing with `jruby-9.0.5.0`, since `jruby-head` has some issues with `bundler` not being installed.

In addition to these changes, I have also enabled Travis's caching feature for Bundler. Hopefully this should speed up our test runs once the caches are primed. In my testing, it appears to have cut the aggregate build time by about a third.


md5:
It's worth noting that Travis is now running tests for three Gemfiles for MRI version 2.2, 2.3, and `head` as well as JRuby 9.0.5.0. The three gemfiles test the default, latest dependencies as specified by the `gemspec`, `rack` 1.x with `json` 1.x, and `rack` 1.x with `json` 2.x.

If we decide to reduce the number of RVM/Gemfile combinations being tested, we can speed up the time needed for Travis to complete.

md5:
Looking closer at the `Gemfile.lock`, it seems that `json` is only actually used as a dependency of `coveralls`, so I think we can safely minimize the combinations tested. I think we do need to test against `rack` 1.x and `rack` 2.x though.

md5:
I've amended this branch to only have `appraisal` gemfiles for `rack` 1 and 2. I've locked the problematic `coveralls` dependencies (`json` and `tins`) in the base `Gemfile`. This should reduce the build matrix by a few builds and speed things up (although not on the first run since it will invalidate the caching due to new gemfile names).

md5:
Looks like this was broken because the build picked up a breaking change between `hashie` 3.4.4 and 3.4.5: https://github.com/intridea/hashie/issues/367. The fix is here: https://github.com/intridea/hashie/pull/368

I think I'll wait until the `hashie` release 3.4.6 is cut and retry this build.

dblock:
I released hashie 3.4.6, please check it out.

md5:
Thanks @dblock! I appreciate the proactive heads-up

md5:
I think this is ready to merge. The Coveralls coverage reported here is not what it shows on the site, probably because I restarted the Travis build without pushing any new commits. I you look at the site, it shows a %2 drop to %94 and the differences look like differences between how coverage is reported across Ruby versions.

supernova32:
Thanks @md5!

PR #114, opened 2016-09-16
Clean up RSpec output

md5:
This PR cleans up the output from RSpec when running on a recent version (e.g. 3.5.0). It makes the following changes:

* Suppress noisy logging
* Move from RSpec's deprecated `should`/`stub` syntax to `expect`/`allow`

This branch is built upon the `appraisal` branch from #113 to make the specs pass. I can rebase once that PR is merged if anyone would like me to do so.


md5:
I'm re-running this after `hashie` 3.4.6 has been pushed.

md5:
Looks like Travis is passing here as well. Rspec output looks much nicer. The Coveralls decrease is %2, same as #113.

md5:
@supernova32 Do you mind taking a look at this one when you get a minute?

supernova32:
Thanks for changing the specs to the `expect` syntax. It looks so much better now. LGTM!

PR #115, opened 2016-09-18
feat: Support for Single Logout

### LINKED ISSUE #81: Support single signout

supernova32:
Replaces #99, as the author there has not responded and we would really like to have this feature for version 1.7.0.

Fixes #81

@md5 can you help me with a review here? I tried solving the conflicts to the best of my abilities, and the tests are green, but I really need more eyes here to make sure the code is sound.

cc @bufferoverflow

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,14 @@ The service provider metadata used to ease configuration of the SAML SP in the I
 * `:idp_sso_target_url` - The URL to which the authentication request should be sent.
   This would be on the identity provider. **Required**.
 
+* `:idp_slo_target_url` - The URL to which the single logout request and response should
+  be sent. This would be on the identity provider. Optional.
+
+* `:slo_default_relay_state` - The value to use as default `RelayState` for single log outs. The
+  value can be a string, or a `Proc` (or other object responding to `call`). The `request`
+  instance will be passed to this callable if it has an arity of 1. If the value is a string,
+  the string will be returned, when the `RelayState` is called. Optional.
+
 * `:idp_sso_target_url_runtime_params` - A dynamic mapping of request params that exist
   during the request phase of OmniAuth that should to be sent to the IdP after a specific
   mapping. So for example, a param `original_request_param` with value `original_param_value`,
@@ -144,6 +152,35 @@ end
 ```
 
 Then follow Devise's general [OmniAuth tutorial](https://github.com/plataformatec/devise/wiki/OmniAuth:-Overview), replacing references to `facebook` with `saml`.
+
+## Single Logout
+
+Single Logout can be Service Provider initiated or Identity Provider initiated.
+When using Devise as an authentication solution, the SP initiated flow can be integrated
+in the `SessionsController#destroy` action.
+
+For this to work it is important to preserve the `saml_uid` value before Devise
+clears the session and redirect to the `/spslo` sub-path to initiate the single logout.
+
+Example `destroy` action in `sessions_controller.rb`:
+
+```ruby
+class SessionsController < Devise::SessionsController
+  # ...
+
+  def destroy
+    # Preserve the saml_uid in the session
+    saml_uid = session["saml_uid"]
+    super do
+      session["saml_uid"] = saml_uid
+      if SAML_SETTINGS.idp_slo_target_url
+        spslo_url = user_omniauth_authorize_url(:saml) + "/spslo"
+        redirect_to(spslo_url)
+      end
+    end
+  end
+end
+```
 
 ## Authors
 
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -27,15 +27,19 @@ module OmniAuth
         first_name: ["first_name", "firstname", "firstName"],
         last_name: ["last_name", "lastname", "lastName"]
       }
+      option :slo_default_relay_state
 
       def request_phase
         options[:assertion_consumer_service_url] ||= callback_url
         runtime_request_parameters = options.delete(:idp_sso_target_url_runtime_params)
 
         additional_params = {}
-        runtime_request_parameters.each_pair do |request_param_key, mapped_param_key|
-          additional_params[mapped_param_key] = request.params[request_param_key.to_s] if request.params.has_key?(request_param_key.to_s)
-        end if runtime_request_parameters
+
+        if runtime_request_parameters
+          runtime_request_parameters.each_pair do |request_param_key, mapped_param_key|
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -86,6 +86,13 @@ The service provider metadata used to ease configuration of the SAML SP in the I
 * `:idp_sso_target_url` - The URL to which the authentication request should be sent.
   This would be on the identity provider. **Required**.
 
+* `:idp_slo_target_url` - The URL to which the single logout request and response should
+  be sent. This would be on the identity provider. Optional.
+
+* `:default_relay_state` - The URL to use as default RelayState on single logouts. The
```
md5: I'm not sure it's accurate to saw that the `RelayState` is a URL in all cases.

Also, can we code-format "RelayState" with backticks?

README.md
```diff
@@ -145,6 +152,32 @@ end
 
 Then follow Devise's general [OmniAuth tutorial](https://github.com/plataformatec/devise/wiki/OmniAuth:-Overview), replacing references to `facebook` with `saml`.
 
+## Single Logout
+
+Single Logout can be Service Provider initiated or Identity Provider initiated.
+The SP initiated flow can be integrated in the `SessionsController#destroy`.
```
md5: This sentence seems to start off assuming that Devise will be used with OmniAuth, which is not always the case. I think it should say something like "When using Devise, the SP-initiated flow can be integrated with the `SessionsController#destroy` action."

README.md
```diff
@@ -145,6 +152,32 @@ end
...
+For this to work its important to preserve the `saml_uid` value before Devise
+clears the session and redirect to the spslo subpath to initiate the single logout.
```
md5: This should say "the `/spslo` subpath" I think.

md5:
@supernova32 I should have been more clear. I think the `README` still needs to be updated, but I think the code is fine.

supernova32:
@md5 don't worry. I did get that 😄 I'll update the README tomorrow.

supernova32:
@md5 I updated the README. Let me know what you think.

README.md
```diff
@@ -86,6 +86,13 @@ The service provider metadata used to ease configuration of the SAML SP in the I
 * `:idp_sso_target_url` - The URL to which the authentication request should be sent.
   This would be on the identity provider. **Required**.
 
+* `:idp_slo_target_url` - The URL to which the single logout request and response should
+  be sent. This would be on the identity provider. Optional.
+
+* `:default_relay_state` - The value to use as default `RelayState` on single log outs. The
```
md5: Now that I read this again, I think it's weird that `:default_relay_state` setting is only used for SLO, despite the fact that the `RelayState` mechanism applies to SSO as well as SLO (depending on the binding). It seems like it should be named `:slo_default_relay_state` or else it should be used in other places as well.

Also, I think this documentation should cover the fact that the `:default_relay_state` (or whatever it is called) can be a `Proc` (or other object responding to `call`) and that the `request` instance will be passed to this callable.

lib/omniauth/strategies/saml.rb
```diff
@@ -146,6 +154,94 @@ def find_attribute_by(keys)
...
+
+        if logout_request.is_valid? &&
+          logout_request.name_id == session["saml_uid"]
+
+          # Actually log out this session
+          session.clear
+
+          # Generate a response to the IdP.
+          logout_request_id = logout_request.id
+          logout_response = OneLogin::RubySaml::SloLogoutresponse.new.create(settings, logout_request_id, nil, RelayState: relay_state)
```
md5: I'm wondering if it makes sense to add the `RelayState` parameter when `relay_state` returns a blank string or `nil`. This applies below to `generate_logout_request` as well.

README.md
```diff
@@ -145,6 +152,35 @@ end
 
 Then follow Devise's general [OmniAuth tutorial](https://github.com/plataformatec/devise/wiki/OmniAuth:-Overview), replacing references to `facebook` with `saml`.
 
+## Single Logout
+
+Single Logout can be Service Provider initiated or Identity Provider initiated.
+When using Devise as authentication solution, the SP initiated flow can be integrated
```
md5: "as an authentication solution"

README.md
```diff
@@ -145,6 +152,35 @@ end
...
+in the `SessionsController#destroy` action.
+
+For this to work its important to preserve the `saml_uid` value before Devise
```
md5: "it's important" or "it is important"

md5:
bump @supernova32

supernova32:
@md5 sorry. I'm a bit swamped at work. I'll take a look at your comments tomorrow.

supernova32:
@md5 i addressed your comments. it's a shame that I cannot resolve a discussion that is no longer valid, or that GitHub does not tell you that that discussion was started on an outdated diff. ¯\\__(ツ)__/¯

supernova32:
@md5 any chance you can sign off on this today? I'd like to cut the new gem over the weekend so we can release the new version with GitLab 8.13. Thanks!

lib/omniauth/strategies/saml.rb
```diff
@@ -44,10 +45,6 @@ def request_phase
       end
 
       def callback_phase
-        unless request.params['SAMLResponse']
-          raise OmniAuth::Strategies::SAML::ValidationError.new("SAML response missing")
```
ErinCall: (or possibly the check should be in `response_fingerprint` itself)

md5: @ErinCall: Nice catch!

Since that's the only call to `response_fingerprint`, it's probably safe to check for `request.params['SAMLResponse']` in that method and return `false` or `nil` if is is not present. This will cause the code to raise `OmniAuth::Strategies::SAML::ValidationError` as it would have previous to this PR.

In general, I think this `idp_cert_fingerprint_validator` code needs to be deprecated and eventually removed, but it's possible I don't understand the use case. I think it's a bad idea to have `ominiauth-saml` directly decoding the `SAMLResponse` instead of relying on `ruby-saml` to do so in a consistent way.

md5:
@supernova32 I'll should be able to spare some time this weekend to review this again if you get the `response_fingerprint` stuff sorted out 👍

supernova32:
@ErinCall thanks for the review. You are absolutely right. I added a guard clause to the `callback_phase` method to raise `OmniAuth::Strategies::SAML::ValidationError` if there is no `SAMLResponse`  present.

@md5 I also cleaned the code up a bit and added some line breaks to make the code easier to read. If you have time today or tomorrow, I'd love to get this merged 😄

Issue #120, opened 2016-11-14
Custom uid in case of transient NameID format


brukau:
At the moment NameID is always used as uid
https://github.com/omniauth/omniauth-saml/blob/master/lib/omniauth/strategies/saml.rb#L139

But according to SAML spec in case of Transient Name Identifier (section 8.3.8, line 3359)
> identifier with transient semantics and SHOULD be treated
> as an opaque and temporary value by the relying party

Will it be possible to add an additional config option to use some attribute instead of NameID?
(Example in djangosaml2 https://github.com/knaperek/djangosaml2/blob/master/djangosaml2/backends.py#L92-L100)

brukau:
Related to https://github.com/omniauth/omniauth-saml/issues/106

md5:
@brukau I'm still happy to review a PR for this change, as I mentioned in #106.

Do you mind if we close this issue? I consider it a duplicate of #106.

md5:
On second thought, I guess it's not exactly a duplicate, but I'd like to see both issues addressed in a single PR.

brukau:
@md5 I would agree with your second comment

md5:
This should have been fixed by #121, though that PR has not yet made it into a release of `omniauth-saml`.

@supernova32 Do you have bandwidth to cut a new release in the near future?

supernova32:
@md5 I have some time tomorrow, I'll cut the release then.

supernova32:
@md5 I just pushed version 1.8.0 of omniauth-saml to GitHub and RubyGems

md5:
:+1:

PR #121, opened 2016-12-01
Add `uid_attribute` option to control the attribute used for the user id

olavmrk:
Some SAML 2.0 IdPs use the transient name identifier format. In that case the name identifier changes for each login, which makes the name identifier unsuitable as a user identifier.

This patch introduces the `uid_attribute` option, that allows us to use a SAML 2.0 attribute as the user identifier instead of the name identifier.

Fixes issue #120.

*Note*: It was suggested in [this issue comment](https://github.com/omniauth/omniauth-saml/issues/120#issuecomment-260744835) that this patch should also make the name identifier optional. I decided to not do that, since that also has implications for logout. I therefore think that makes more sense as a separate patch.

### DIFF
--- a/spec/omniauth/strategies/saml_spec.rb
+++ b/spec/omniauth/strategies/saml_spec.rb
@@ -207,6 +207,31 @@ describe OmniAuth::Strategies::SAML, :type => :strategy do
       end
     end
 
+    context "when using custom user id attribute" do
+      before :each do
+        saml_options[:idp_cert_fingerprint] = "3B:82:F1:F5:54:FC:A8:FF:12:B8:4B:B8:16:61:1D:E4:8E:9B:E2:3C"
+        saml_options[:uid_attribute] = "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress"
+        post_xml :custom_attributes
+      end
+
+      it "should return user id attribute" do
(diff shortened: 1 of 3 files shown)


spec/omniauth/strategies/saml_spec.rb
```diff
@@ -207,6 +207,29 @@ def post_xml(xml=:example_response, opts = {})
...
+    end
+
+    context "when using custom user id attribute, but it is missing" do
+      before :each do
+        saml_options[:uid_attribute] = "missing_attribute"
+        post_xml
+      end
+
+      it "should not return an authenticated session" do
+        expect(auth_hash).to eq(nil)
```
md5: I'm confused about why `auth_hash` is `nil` instead of the `ValidationError` being raised. Does the error get rescued somewhere in the OmniAuth code?

olavmrk: That was what I had assumed, but upon a closer look, I realized that it was actually handled by [this exception handler in `strategies/saml.rb`](https://github.com/omniauth/omniauth-saml/blob/v1.7.0/lib/omniauth/strategies/saml.rb#L79).

With that in mind I see how to add a more comprehensive test for this error, so I have now implemented that.

olavmrk:
@md5: I have updated the commit message now. Sorry for not noticing the contributor guidelines earlier.

PR #123, opened 2016-12-21
Add more detail to the readme on SSO

iainbeeston:
Currently the readme doesn't have a lot of detail on SSO. I've recently had to implement SSO and it required a lot of reading into the source code for both omniauth-saml and ruby-saml. I've added what I have learned into the readme, to help anyone who needs to configure this in the future. I'd really appreciate a technical review from someone else who is familiar with this, in case I'm not clear or have made a mistake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,12 @@ Then follow Devise's general [OmniAuth tutorial](https://github.com/plataformate
 ## Single Logout
 
 Single Logout can be Service Provider initiated or Identity Provider initiated.
+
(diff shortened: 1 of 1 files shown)


iainbeeston:
Thanks @md5 - I've changed some more of this now, as devise throws an error if you explicitly redirect inside a controller action (instead of using one of the redirect-after-action callback methods)

iainbeeston:
@md5 Sorry for the delay - I've squashed this now as requested. I think it should be ready to merge(?)

PR #124, opened 2017-01-17
feat: include SessionIndex in logout requests

iainbeeston:
Some idp's use the sessionindex rather than the uid to find the correct
session to sign out. I've made this work by saving it in the session
alongside the uid, then putting it back into the logout request where
ruby-saml can handle it, and include it in the logout request.

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -159,7 +159,7 @@ module OmniAuth
         Hash[found_attributes]
       end
 
-      extra { { :raw_info => @attributes, :response_object =>  @response_object } }
+      extra { { :raw_info => @attributes, :session_index => @session_index, :response_object =>  @response_object } }
 
       def find_attribute_by(keys)
         keys.each do |key|
@@ -182,6 +182,7 @@ module OmniAuth
 
         response.is_valid?
         @name_id = response.name_id
+        @session_index = response.sessionindex
         @attributes = response.attributes
         @response_object = response
 
@@ -190,6 +191,7 @@ module OmniAuth
         end
 
         session["saml_uid"] = @name_id
+        session["saml_session_index"] = @session_index
         yield
       end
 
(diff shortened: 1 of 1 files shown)


iainbeeston:
The build error seems to be because the latest of version of nokogiri requires ruby 2.1+

md5:
@iainbeeston Right. We can fix this by updating the `Appraisals` file and `.travis.yml`, but my concern is that we already have a pretty large test matrix and I don't want to inflate it by having another axis to consider.

I would honestly love to drop Ruby 1.9 and 2.0 support from this gem entirely, but since `omniauth` itself and `ruby-saml` still support the older Ruby versions, I don't that's actually a reasonable move to make.

Another possibility would be to repurpose the "rack-1" appraisal to simply be the "legacy" gems appraisal and to add a `gem 'nokogiri', '~> 1.6.8'` constraint there.

@omniauth/omniauth-saml-maintainers Any thoughts on this?

iainbeeston:
@md5 Is it worth opening a separate issue for the build/ruby version problem?

supernova32:
@md5 I like your idea of repurposing the `rack-1` appraisal to be the "legacy" gem appraisal.

otherwise this PR looks 👍

fenicks:
Is this pullrequest going to next version? I'm working with Shibboleth v3.2.1 and for the SLO SessionIndex is mandatory. I worked exactly on the same thing: [https://github.com/fenicks/omniauth-saml/tree/feature-sessionindex](https://github.com/fenicks/omniauth-saml/tree/feature-sessionindex).

iainbeeston:
Is there any chance of getting this merged? Anything I can do?

md5:
@iainbeeston I've opened https://github.com/omniauth/omniauth-saml/pull/130 to remove support for Ruby 1.9 and 2.0. Hopefully we can get that merged in a timely fashion and you can merge `master` into your branch to get this PR moving again.

iainbeeston:
Great, thanks for the update!

supernova32:
@iainbeeston #130 was merged. You can go ahead and rebase your branch against master, so that we can merge it.

iainbeeston:
@supernova32 I've rebased now. Hopefully this should be ready to merge!

PR #130, opened 2017-03-22
Require Ruby 2.1+

### LINKED ISSUE #126: Drop support for Ruby 1.9 and 2.0
It looks like `omniauth` has now dropped support for EOL versions of Ruby as of version 1.5.0. I think we should follow suit and only support Ruby 2.1 or higher. This should greatly simplify the Travis config.

cf. omniauth/omniauth#883

md5:
Fixes #126

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,6 @@ script:
   - bundle exec rspec
 
 rvm:
-  - 1.9.3
-  - jruby-19mode
-  - 2.0.0
   - 2.1.10
   - 2.2.2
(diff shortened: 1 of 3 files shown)


md5:
@supernova32 Thanks for the review. I amended my commit to include an update to `required_ruby_version` in the `*.gemspec` as well. This will let RubyGems.org know about the new version requirement. See https://github.com/omniauth/omniauth-saml/pull/130/commits/009c7c625beb89e736799cfdcc56c85204fc1064

PR #134, opened 2017-05-03
feat: Add option :settings_object

tosch:
This allows using the `OneLogin::RubySaml::IdpMetadataParser` for
configuration.

Other options given to the strategy will be merged into that settings
object. If `:settings_object` is `nil`, which is the default, a new
instance is created using the options given to the strategy.


tosch:
This is related to #83

tosch:
I'm closing that for now, this doesn't work as it should. The `settings_object` gets modified during single phase calls, this yields incorrect results when fetching the metadata for instance. I hope to come up with another solution to allow parsing the IdP metadata in the next days.

Issue #112, opened 2016-09-15
Support multiple signing keys


mceldeen:
According to section 2.4.1.1 of [Metadata for the OASIS Security Assertion Markup Language (SAML) V2.0 spec](https://www.oasis-open.org/committees/download.php/35391/sstc-saml-metadata-errata-2.0-wd-04-diff.pdf), multiple signing certs may be provided for a given IdP. This provider does not currently support this functionality. One reason this matters is because AzureAD via SAML 2.0 happens to use multiple signing certs which makes this gem incompatible with AzureAD.

This change would [require a change to the ruby-saml gem](https://github.com/onelogin/ruby-saml/issues/329) so this cannot be immediately implemented using the `ruby-saml` gem. However,  the discussion in the relevant pull request indicates that those changes will be made relatively soon (perhaps end of September 2016).

md5:
This is now possible by upgrading to `ruby-saml` 1.4.3. Information on how to use multiple IdP certificates can be found in the `ruby-saml` documentation. In addition, information on how to configure the IdP certificates from an XML metadata file was added to the `README` in #135.

PR #135, opened 2017-05-08
Chore: Add section about parsing IdP metadata to README

### LINKED ISSUE #83: Add support for parsing remote IdP metadata

tosch:
***Should not be merged before onelogin/ruby-saml#393 is in their master***

Fixes #83

### DIFF
--- a/omniauth-saml.gemspec
+++ b/omniauth-saml.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |gem|
   gem.required_ruby_version = '>= 2.1'
 
   gem.add_runtime_dependency 'omniauth', '~> 1.3'
-  gem.add_runtime_dependency 'ruby-saml', '~> 1.4'
+  gem.add_runtime_dependency 'ruby-saml', '~> 1.4', '>= 1.4.3'
 
   gem.add_development_dependency 'rake', '>= 10', '< 12'
   gem.add_development_dependency 'rspec', '~>3.4'
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ For IdP-initiated SSO, users should directly access the IdP SSO target URL. Set 
 
 A `OneLogin::RubySaml::Response` object is added to the `env['omniauth.auth']` extra attribute, so we can use it in the controller via `env['omniauth.auth'].extra.response_object`
 
-## Metadata
+## SP Metadata
 
 The service provider metadata used to ease configuration of the SAML SP in the IdP can be retrieved from `http://example.com/auth/saml/metadata`. Send this URL to the administrator of the IdP.
 
@@ -144,6 +144,27 @@ Note that when [integrating with Devise](#devise-integration), the URL path will
 * `:uid_attribute` - Attribute that uniquely identifies the user. If unset, the name identifier returned by the IdP is used.
 
 * See the `OneLogin::RubySaml::Settings` class in the [Ruby SAML gem](https://github.com/onelogin/ruby-saml) for additional supported options.
+
+## IdP Metadata
+
+You can use the `OneLogin::RubySaml::IdpMetadataParser` to configure some options:
+
+```ruby
+require 'omniauth'
+idp_metadata_parser = OneLogin::RubySaml::IdpMetadataParser.new
+idp_metadata = idp_metadata_parser.parse_remote_to_hash("http://idp.example.com/saml/metadata")
+
+# or, if you have the metadata in a String:
+# idp_metadata = idp_metadata_parser.parse_to_hash(idp_metadata_xml)
+
+use OmniAuth::Strategies::SAML,
+  idp_metadata.merge(
+    :assertion_consumer_service_url => "consumer_service_url",
+    :issuer                         => "issuer"
+  )
+```
+
+See the [Ruby SAML gem's README](https://github.com/onelogin/ruby-saml#metadata-based-configuration) for more details.
 
 ## Devise Integration
 


md5:
@tosch Have you had a look at https://github.com/onelogin/ruby-saml/pull/389? I think it would be good to release an update to `ruby-saml` to support having multiple IDP keys at the same time this PR is released. See #112

md5:
After spending some time looking at this PR, it seems pretty complicated for the value it's providing.

For one, I don't think that mixing the phase method refactoring with the support for IDP metadata parsing in a single PR makes sense, but perhaps I didn't look closely enough at the changes.

Second, I don't see much reason to specifically support fetching metadata from a URL. An application could just as easily have its IdP metadata in a database, in a file, or somewhere else.

Lastly, I'm not sure I understand the use case for overriding metadata values retrieved from the IdP. Based on my reading of [`IdpMetadataParser#parse`](https://github.com/onelogin/ruby-saml/blob/39935bd/lib/onelogin/ruby-saml/idp_metadata_parser.rb#L42) from `ruby-saml`, the only values it sets are:

* `:idp_entity_id`
* `:name_identifier_format`
* `:idp_sso_target_url`
* `:idp_slo_target_url`
* `:idp_cert`
* `:idp_cert_fingerprint`
* `:idp_attribute_names`

Could you explain when you would want to override any of those values?

Even without this patch, it's currently not that hard to configure from IdP metadata. Here is an example adapted from the `omniauth-saml` docs:

```ruby
idp_options = OneLogin::RubySaml::IdpMetadataParser.new.parse(
  File.new(File.expand_path('../config/idp_metadata.xml', __FILE__)).read
)

use OmniAuth::Strategies::SAML,
    idp_options.merge(
      assertion_consumer_service_url: "consumer_service_url",
      issuer: "issuer"
    )
```

You could just as easily have code that fetches the metadata from a URL in your initialization code and it would only be run once without having to do things like override the `OmniAuth::Strategy` constructor.

Assuming I'm not missing something, I think we could be OK just documenting this pattern instead of adding special support for it.

tosch:
@md5 Thanks for your review. I totally agree and will change the PR to only update the Readme (and open up another one for the refactoring commit?)

tosch:
The only issue with your example is that the IdpMetadataParser doesn't return an Hash, but a Settings object which does not respond to `#merge`. It is probably better to have a `#to_hash` method on `OneLogin::RubySaml::Settings`, though, instead of having the glue code here.

md5:
I guess I was mixing up the `ruby-saml` settings object with the `omniauth` one, the latter being a `Hashie::Mash` subclass.

md5:
@pitbulk would you guys accept a patch to allow `OneLogin::RubySaml::Settings` to respond to `merge` and/or `to_h`?

tosch:
@pitbulk Or a `#parse_to_hash` method for the `IdpMetadataParser`. This would not break the existing interface, but allow using the parser together with omniauth-saml

md5:
Having a separate `to_h` seems more composable to me and easier to test in `ruby-saml` itself. I don't think adding a method or two breaks the existing interface, it just expands it (as would adding your proposed `parse_to_hash`).

tosch:
That's true, of course. My reasoning for prefering `IdpMetadataParser#parse_to_hash` over `Settings#to_hash` is that the first method would only return the settings provided by the metadata, while the second one would also return all the ruby-saml default settings. The user would have to be very careful about the order in which he merges the different settings to not overwrite his own values with the default ones.

tosch:
To show this in code:

```ruby
parser = OneLogin::RubySaml::IdpMetadataParser.new
metadata_xml = File.read('/some/path.xml')

idp_settings_hash = parser.parse_to_hash(metadata_xml)
# idp_settings_hash is now:
# {
#   idp_entity_id: 'abc',
#   name_identifier_format: 'foo',
#   idp_sso_target_url: 'https://some.where/saml/sso',
#   idp_slo_target_url: 'https://some.where/saml/slo',
#   idp_cert: 'dfdf',
#   idp_cert_fingerprint: 'ab:cd',
#   idp_attribute_names: []
# }

all_settings_hash = parse.parse(metadata_xml)
# all_settings_hash is now:
# {
#   :single_logout_service_binding => "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect",
#   :idp_entity_id => "https://some.where/saml2/idp/metadata.php",
#   :idp_sso_target_url => "https://some.where/saml2/idp/SSOService.php",
#   :idp_slo_target_url => "https://some.where/saml2/idp/SingleLogoutService.php",
#   :idp_cert => 'dfdf',
#   :idp_cert_fingerprint_algorithm=>"http://www.w3.org/2000/09/xmldsig#sha1",
#   :idp_attribute_names=>[],
#   :assertion_consumer_service_binding=>"urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST",
#   :double_quote_xml_attribute_values=>false,
#   :name_identifier_format=>"urn:oasis:names:tc:SAML:2.0:nameid-format:transient",
#   :compress_response=>true,
#   :idp_cert_fingerprint=>"aa:bb",
#   :compress_request=>true,
#   :security=> {
#     :authn_requests_signed=>false,
#     :logout_requests_signed=>false,
#     :logout_responses_signed=>false,
#     :want_assertions_signed=>false,
#     :want_assertions_encrypted=>false,
#     :want_name_id=>false,
#     :metadata_signed=>false,
#     :embed_sign=>false,
#     :digest_method=>"http://www.w3.org/2000/09/xmldsig#sha1",
#     :signature_method"=>"http://www.w3.org/2000/09/xmldsig#rsa-sha1"
#   },
#   :soft=>true
# }
```

md5:
Thanks @tosch. I see what you mean. I think my main concern was that if you introduce `parse_to_hash`, do you also need to add `parse_remote_to_hash`?

Looking closer at the `OmniAuth::Strategy` constructor, it looks like it pretty strictly requires a `Hash`, so passing `OneLogin::RubySaml::Settings` wouldn't work at all.

Let's see what @pitbulk has to say about all this since what we're talking about is changes to `ruby-saml`.

pitbulk:
The ruby-saml toolkit itself doesn't need the parse_to_hash / parse_remote_to_hash methods, but if you guys need it here, I'm ok adding them.

@tosch do you want to contribute them to the [key_rollover_mngmt](https://github.com/onelogin/ruby-saml/tree/key_rollover_mngmt) branch, so I will be able to integrate them on the [PR](https://github.com/onelogin/ruby-saml/pull/389)?

tosch:
@pitbulk Will do within the next few days.

tosch:
Updated the PR to only add a new `IdP Metadata` section to the Readme after creating a PR for ruby-saml

Thanks for all the feedback :)

md5:
@tosch If you want to bump the `ruby-saml` version in this PR as well, I'd be fine with that. Otherwise, we can do a separate PR to bump to a version that supports `parse_to_hash`.

tosch:
@md5 I've pushed a commit that bumps the version of ruby-saml

omniauth-saml.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |gem|
   gem.required_ruby_version = '>= 2.1'
 
   gem.add_runtime_dependency 'omniauth', '~> 1.3'
-  gem.add_runtime_dependency 'ruby-saml', '~> 1.4'
+  gem.add_runtime_dependency 'ruby-saml', '~> 1.4', '>= 1.4.3'
```
md5: I find it odd that this wasn't a 1.5 release, but it is what it is.

PR #136, opened 2017-05-09
refactor: Split up phase methods

tosch:
This encapsulates different aspects of the business logic better and allows overwriting certain aspects of it when desired.

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -33,51 +33,25 @@ module OmniAuth
 
       def request_phase
         options[:assertion_consumer_service_url] ||= callback_url
-        runtime_request_parameters = options.delete(:idp_sso_target_url_runtime_params)
-
-        additional_params = {}
-
-        if runtime_request_parameters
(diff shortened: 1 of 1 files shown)


tosch:
Not sure why the coveralls status didn't update, but it ran successfully.

md5:
This PR looks reasonable to me on the whole. The one thing that slightly concerns me is the increase in the number of public methods.

Making `other_phase_for_*` public makes sense to me since that's the intent of this PR, but I wonder about making methods like `validate_fingerprint`, `with_settings`, and `add_request_attributes_to` public. Perhaps this PR could make them private and they can be exposed later if there's a demonstrable need to override their behavior? Along the same lines, perhaps the `other_phase_for_*` method should actually be protected instead of public. This would allow subclasses to override them without increasing the public API of the strategy.

@supernova32 Do you have any thoughts on this?

tosch:
@md5 As far as I can see, those methods are private, they are defined below that line: https://github.com/omniauth/omniauth-saml/pull/136/files#diff-41b1e8a1e3bbb14a039661f1b0c755bcR123

md5:
@tosch Argh! I'll look more carefully next time

PR #137, opened 2017-05-09
Chore: Increase Coverage

tosch:
While extending the specs to increase the coverage, I noticed that `OmniAuth::Strategies::SAML::ValidationError` exceptions are not handled in the `other_request` phase. This will result in returning `500 Server Error` responses, while some `4xx` error would be probably be more appropriate.

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -186,10 +186,6 @@ module OmniAuth
         @attributes = response.attributes
         @response_object = response
 
-        if @name_id.nil? || @name_id.empty?
-          raise OmniAuth::Strategies::SAML::ValidationError.new("SAML response missing 'name_id'")
-        end
-
         session["saml_uid"] = @name_id
         session["saml_session_index"] = @session_index
         yield
--- a/spec/omniauth/strategies/saml_spec.rb
+++ b/spec/omniauth/strategies/saml_spec.rb
@@ -125,22 +125,35 @@ describe OmniAuth::Strategies::SAML, :type => :strategy do
     context "when fingerprint is empty and there's a fingerprint validator" do
       before :each do
         saml_options.delete(:idp_cert_fingerprint)
-        saml_options[:idp_cert_fingerprint_validator] = lambda { |fingerprint| "C1:59:74:2B:E8:0C:6C:A9:41:0F:6E:83:F6:D1:52:25:45:58:89:FB" }
-        post_xml
+        saml_options[:idp_cert_fingerprint_validator] = fingerprint_validator
       end
 
-      it "should set the uid to the nameID in the SAML response" do
-        expect(auth_hash['uid']).to eq '_1f6fcf6be5e13b08b1e3610e7ff59f205fbd814f23'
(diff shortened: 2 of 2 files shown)


spec/omniauth/strategies/saml_spec.rb
```diff
@@ -10,6 +10,15 @@ def post_xml(xml=:example_response, opts = {})
   post "/auth/saml/callback", opts.merge({'SAMLResponse' => load_xml(xml)})
 end
 
+def post_logout(params = nil)
```
md5: How about making this the `subject` of a new context block that surrounds the three context blocks that use this method instead? I think it's easier to follow the specs if this definition is near the examples it supports. I'm not much of a fan of the existing `post_xml` method for that reason.

tosch:
@md5 Addressed your comment.

Another question: Should I create an issue for the exceptions not being handled in the `other_phase`?

md5:
@tosch Yes, please create an issue.

spec/omniauth/strategies/saml_spec.rb
```diff
@@ -125,22 +125,35 @@ def post_xml(xml=:example_response, opts = {})
...
-          'email'        => ['user@example.com'],
-          'company_name' => ['Example Company'],
-          'fingerprint'  => 'C1:59:74:2B:E8:0C:6C:A9:41:0F:6E:83:F6:D1:52:25:45:58:89:FB'
-        )
+      context "when the fingerprint validator returns false" do
+        let(:fingerprint_validator) { lambda { |_| false } }
+
+        before { post_xml }
+
+        it { should fail_with(:invalid_ticket) }
```
md5: I just noticed we're using a mix of expect and should syntax. Can you make this `is_expected.to`?

lib/omniauth/strategies/saml.rb
```diff
@@ -186,9 +186,12 @@ def handle_response(raw_response, opts, settings)
         @attributes = response.attributes
         @response_object = response
 
+        # With ruby-saml 1.4.2, that line is never reached with an empty name_id on the response
+        # :nocov:
         if @name_id.nil? || @name_id.empty?
           raise OmniAuth::Strategies::SAML::ValidationError.new("SAML response missing 'name_id'")
         end
+        # :nocov:
```
supernova32: I'm not a fan of these type of comments, that really have nothing to do with the code itself, but are just a way to achieve 100% coverage. See [here](https://dev.to/danlebrero/the-tragedy-of-100-code-coverage) for why trying to always achieve 100% coverage can be a bad thing.

I would remove them.

tosch: Actually, I think this code can be removed. The ruby-saml is required to be `~> 1.4` and the oldest version matching that tests for the `name_id` presence itself.

Wdyt?

supernova32: @tosch sure, if the presence of `name_id` is already checked in `ruby-saml` then there is no need to check for it here as well.

tosch:
@md5 @supernova32 I've addressed your comments 😄  Thanks for your reviews.

PR #140, opened 2017-06-08
fix: default assertion_consumer_service_url not set during callback

### LINKED ISSUE #139: ruby-saml 1.4.3 breaks default assertion_consumer_service_url
ruby-saml 1.4.3 added SubjectConfirmation Recipient validation. This means that assertion_consumer_service_url must be set in the ruby-saml settings during the callback_phase. Typically I depend on omniauth-saml to set this automatically. However, the default is not set during the callback phase, which causes a SubjectConfirmation Recipient validation failure.

Moving the line

```
options[:assertion_consumer_service_url] ||= callback_url
```

into `with_settings` seems to fix the issue. I was about to create a PR, but had trouble creating a spec since it it uses a Recipient of localhost, but a callback_url of example.org.

justinhoward:
Fix a bug where ruby-saml would fail SubjectConfirmation Recipient
validation when using the default assertion_consumer_service_url. The
url was not being set during the callback phase.

This closes #139

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -32,8 +32,6 @@ module OmniAuth
       option :idp_slo_session_destroy, proc { |_env, session| session.clear }
 
       def request_phase
-        options[:assertion_consumer_service_url] ||= callback_url
-
         authn_request = OneLogin::RubySaml::Authrequest.new
 
         with_settings do |settings|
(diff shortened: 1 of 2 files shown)


tmertens:
Can we get a new gem release including this fix?  Thanks!

wingrunr21:
@md5 @supernova32 

without this fix in a published version of the gem there are some incompatibilities with the README. One needs `ruby-saml` >= 1.4.3. to get the `parse_remote_to_hash` method as referenced, but that version also causes the problem addressed in this PR. Mind pushing a version bump?

supernova32:
@tmertens @wingrunr21 Version 1.8.1 has been released.

wingrunr21:
Thanks!

tmertens:
Thank you!

PR #147, opened 2017-11-27
Upgrade ruby-saml to 1.6.0

SokratisVidros:


md5:
@SokratisVidros Can you provide more background on why this is update is necessary?

SokratisVidros:
@md5 the main reason is to leverage the latest version of validate_destination of ruby-saml gem. https://github.com/onelogin/ruby-saml/blob/v1.6.0/lib/onelogin/ruby-saml/response.rb#L612. We could have also used v1.5.0 but if there are no breaking changes why not migrate to v1.6.0 at first place.

SokratisVidros:
@md5 any updates on this one?

md5:
I still don't see any reason to force an upgrade.

bufferoverflow:
there are a few bug fixes with 1.5.0 and 1.6.0, see https://github.com/onelogin/ruby-saml/blob/master/changelog.md , however some adoptions might become required, https://github.com/onelogin/ruby-saml#updating-from-150-to-160

SokratisVidros:
@bufferoverflow, I've been using the patched version for two months with multiple IdP vendors and I haven't noticed any discrepancies. I guess that before the upgrade, extensive testing especially with Azure ADFS is required. I will get back to you as soon as we have the results.

SokratisVidros:
It seems like the upgrade changes logout requests initialisation. More specifically, params Signature, SAMLRequest, SigAlg and RelayState should be passed explicitly on specific options before they get URI decoded. According to the source code, this addresses the following:

```
Further, note that URL-encoding is not canonical; that is, there are multiple legal encodings for a given value. The relying party MUST therefore perform the verification step using the original URL-encoded values it received on the query string. It is not sufficient to re-encode the parameters after they have been processed by software because the resulting encoding may not match the signer's encoding.
```

https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/slo_logoutrequest.rb#L247

However, omniauth-saml gem doesn't verify signatures on SLO requests. As you can see, the constructor expects the params on a second argument (options) that is never passed.

https://github.com/omniauth/omniauth-saml/blob/master/lib/omniauth/strategies/saml.rb#L176

All in all, it seems like the upgrade is not affected by ruby-saml changes. I hope that helps.

tommymccallig:
For what it's worth, we've recently run into an issue which could have been diagnosed using the improved error handling in https://github.com/onelogin/ruby-saml/issues/374 ( `ruby-saml#1.5.0`). 

It includes extra information as to why an IDP returned a non-success code.

md5:
I've opened an alternative PR at https://github.com/omniauth/omniauth-saml/pull/155 that updates the SLO code to pass the settings objects and raw params as recommended. It would be great if we could get some testing, particularly from ADFS users who are using an encoding other than UTF-8.

md5:
Fixed by #157

PR #154, opened 2018-02-19
fix: ambiguous path match in other phase

### LINKED ISSUE #153: Matching wrong provider when name is a substring of another provider
Because the check in `other_phase` uses `current_path.start_with?(request_path)`, it will match a provider whose name is a substring of another provider.

https://github.com/omniauth/omniauth-saml/blob/master/lib/omniauth/strategies/saml.rb#L72

`request_path` is defined in https://github.com/omniauth/omniauth/blob/master/lib/omniauth/strategy.rb#L386
as 
```ruby
@request_path ||= options[:request_path].is_a?(String) ?
    options[:request_path] :
    "#{path_prefix}/#{name}"
```

This can result in the wrong provider being set in `env['omniauth.strategy']` and the wrong setup phase being run, etc.

Example:

`provider1`
`provider10`

If the path is `/auth/provider10/metadata`, then we'll match `provider1`, set `env['omniauth.strategy']` to `provider1` and run its setup phase.

tommymccallig:
This closes #153

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -69,7 +69,7 @@ module OmniAuth
       end
 
       def other_phase
-        if current_path.start_with?(request_path)
+        if request_path_pattern.match(current_path)
           @env['omniauth.strategy'] ||= self
           setup_phase
 
@@ -119,6 +119,10 @@ module OmniAuth
       end
 
       private
+
+      def request_path_pattern
+        @request_path_pattern ||= %r{\A#{Regexp.quote(request_path)}(/|\z)}
+      end
 
       def on_subpath?(subpath)
         on_path?("#{request_path}/#{subpath}")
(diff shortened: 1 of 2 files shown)


lib/omniauth/strategies/saml.rb
```diff
@@ -69,7 +69,7 @@ def response_fingerprint
       end
 
       def other_phase
-        if current_path.start_with?(request_path)
+        if current_path.match(/#{Regexp.quote(request_path)}[\/\z]{1}/)
```
md5: There are a few problems with this regexp:

1. It needs to be anchored at the beginning. It currently can math things that have a prefix in front of `request_path`
2. I'm pretty sure `\z` doesn't work inside a character class. You need to use `(/|\z)`
3. The `{1}` isn't doing anything.

Since the underlying implementation of `request_path` memoizes its value and doesn't change over the lifecycle of the strategy instance, I think I would introduce a method like this:

```ruby
def request_path_pattern
  @request_path_pattern ||= /\A#{Regexp.quote(request_path)}(/|\z)/
end
```

Then, in `other_phase`, I would call `if request_path_pattern.match(current_path)`.

md5: This will still break if someone uses `options[:request_path]` and puts a trailing slash on it, but it's better than the current situation.

tommymccallig:
@md5 Thanks for the feedback! Changes made and rebased

PR #157, opened 2018-02-27
Fix CVE-2017-11430

md5:
Fixes #156

PR #158, opened 2018-03-01
Backport ruby-saml bump

MrBerg:
CVE-2017-11430 seems serious enough to warrant a backport and patch for people who have not yet updated to Ruby ~> 2.0 (or JRuby 9k). So this is essentially a repeat of https://github.com/omniauth/omniauth-saml/pull/157, but for 1.7.0

(This should have been based on v1.7.0 but it seems I don't have that option)


supernova32:
@MrBerg thanks for this, but we cannot merge this as is. It should be based of v1.7.0, but since we don't have version branches, that makes it a bit difficult. I guess the way forward would be to checkout the v1.7.0 tag, create a branch for it called `1-7-stable` and push it to origin. Once that branch is ready, we can merge these changes there and cut a new release.

@md5 thoughts?

md5:
Given that the fix is actually to upgrade `ruby-saml` and there’s nothing preventing that with version 1.7.0, I’m not convinced we need to backport.

I also don’t think we should cut a branch to maintain a version that’s compatible with EOL'd Ruby versions.

supernova32:
That's a good reasoning. I totally agree. Thanks @md5

PR #159, opened 2018-03-01
Whitelist more OneLogin::RubySaml::Response options

mberlanda:
I recently updated this gem in a repository where it was in use `v1.4.2`.
I noticed that https://github.com/omniauth/omniauth-saml/commit/4712e9990dbcb768286941d2f7f50e1b0e048d49 introduced a whitelist logic for options passed to the strategy:

```rb
# lib/omniauth/strategies/saml.rb
OTHER_REQUEST_OPTIONS = [:skip_conditions, :allowed_clock_drift, :matches_request_id, :skip_subject_confirmation].freeze
``` 

To this list, however, is excluding several options accepted by `OneLogin::RubySaml::Response`. Thus I added a couple of options that were missing and are used in [`ruby-saml`](https://github.com/onelogin/ruby-saml):

- `:skip_destination` https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/response.rb#L593-L615
- `: skip_recipient_check ` https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/response.rb#L744

PR #165, opened 2018-05-23
Retrieves response available options from ruby-saml

mberlanda:
In https://github.com/onelogin/ruby-saml/pull/454 was introduced a whitelisting for allowed Response options.

In my previous pr https://github.com/omniauth/omniauth-saml/pull/159 I added one more option.
According to me, this whitelist introduced in 4712e9990dbcb768286941d2f7f50e1b0e048d49 should not be a responsibility of `omniauth-saml` but it should be provided out of the box by `ruby-saml`.  It looked like @md5 agreed on this point (https://github.com/omniauth/omniauth-saml/pull/159#issuecomment-376233630)

Since `'ruby-saml' ~>1.8` integrated this change, I updated the dependency of this gem and retrieved response available options from it.

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -10,10 +10,7 @@ module OmniAuth
         OmniAuth::Strategy.included(subclass)
       end
 
-      OTHER_REQUEST_OPTIONS = [
-        :skip_conditions, :allowed_clock_drift, :matches_request_id,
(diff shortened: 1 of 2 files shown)


mberlanda:
hey @md5 @supernova32 , do you have any feedback on this pr?

supernova32:
@mberlanda thanks for the ping, and thanks for the PR. I think it looks good, and I'm inclined to merge it. I'll wait until tomorrow, so that if @md5, or any one else has any objections, they can let us know.

PR #169, opened 2018-09-04
Update ruby-saml dependency 1.9.0 version

SokratisVidros:
Changes for v1.8.x and 1.9.x can be found at https://github.com/onelogin/ruby-saml/blob/master/changelog.md

Thanks,
Sokratis

### DIFF
--- a/omniauth-saml.gemspec
+++ b/omniauth-saml.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |gem|
   gem.required_ruby_version = '>= 2.1'
 
   gem.add_runtime_dependency 'omniauth', '~> 1.3', '>= 1.3.2'
-  gem.add_runtime_dependency 'ruby-saml', '~> 1.8'
+  gem.add_runtime_dependency 'ruby-saml', '~> 1.9'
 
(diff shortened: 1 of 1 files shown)


md5:
@SokratisVidros What's the need to upgrade this dependency? I don't see any security fixes in the changelog for 1.8 and 1.9, so I'm not sure we have a reason to bump this. Users can already install the latest version of `ruby-saml` if they wish.

SokratisVidros:
We use omniauth-saml in our app however we need to be able to get the latest change from ruby-saml 1.9 regarding the "allowed_clock_drift" setting in non number format. It would be really helpful to avoid forking this repo in the meantime.

md5:
@SokratisVidros There is no need to fork. 1.9 is compatible with the `~> 1.7` version specification.

supernova32:
I agree with @md5 here. Since there are no vulnerability fixes between the versions, I see little gain by increasing the required minimum version of `ruby-saml`. As Mike mentioned, you can easily add `ruby-saml` 1.9 to your Gemfile and the configuration options will work without having to fork.

Issue #172, opened 2018-11-08
Ability to store `authn_request.uuid` for InResponseTo comparison


Jamedjo:
I'd like to verify that certain requests were initiated from the service provider, rather than being unsolicited ones from the IdP. I'd like to do this by storing `authn_request.uuid` from `#request_phase` and then later comparing this to `InResponseTo`. This might involve `matches_request_id`, or might bypass that to sometimes allow unsolicited IdP initiated requests.

The [SAML protocols spec](https://www.oasis-open.org/committees/download.php/35389/sstc-saml-profiles-errata-2.0-wd-06-diff.pdf) section on `4.1.4 Use of Authentication Request Protocol` includes the following:

> 4.1.4.3 <Response> Message Processing Rules
> Regardless of the SAML binding used, the service provider MUST do the following:
> • ...
> • Verify that the `InResponseTo` attribute in the bearer `<SubjectConfirmationData>` equals the ID of its original `<AuthnRequest>` message, unless the response is unsolicited (see Section 4.1.5), in which case the attribute MUST NOT be present

Unfortunately I don't have any way to access `authn_request.uuid` during the request phase to make this work.

PR #174, opened 2019-03-25
feat: allow request uuid to be stored

Jamedjo:
## What

Introduces a :store_request_uuid option for later comparison with InResponseTo

By default it saves the request uuid in the session as "saml_transaction_id",
but also accepts a proc that will then be called with the uuid for custom storage.

## Why

Needed for #172, although we may also want to pass the value to ruby-saml with `matches_request_id:`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,10 @@ be present.
 
 * `:uid_attribute` - Attribute that uniquely identifies the user. If unset, the name identifier returned by the IdP is used.
 
+* `:store_request_uuid` - Used to store the request's UUID for later verification of InReponseTo.
+  By default it saves the request uuid in the session as "saml_transaction_id",
+  but also accepts a proc that will then be called with the uuid for custom storage.
+
 * See the `OneLogin::RubySaml::Settings` class in the [Ruby SAML gem](https://github.com/onelogin/ruby-saml) for additional supported options.
 
 ## IdP Metadata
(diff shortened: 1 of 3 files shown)


Jamedjo:
@md5 @supernova32 Does this look ok?

suprnova32 (dismissed):
@Jamedjo the code looks good. Are you using this already in your apps?

bufferoverflow (dismissed):

alexrecuenco:
Was this ever solved in a different way? I see no updates here, and I was trying to do SP-initiated only log-in by looking at the InResponseTo, but I don't think that is currently possible, is it?

Is there any recommendation to avoid CSFR otherwise?

How do you recommend to go about this?

alexrecuenco:
I just want to mention, if you are reading this 6+ years later, because you are using omniauth-saml and trying to do a SP flow by implementing a patch like this.

If you use `SameSite` `Lax` or `Strict` on your session cookies, which you probably should do, it won't work because the reply back is a POST request, so it doesn't have any session cookie.

Use a separate cookie/session [signed and/or encrypted] to store the session id, possibly using `__Host-` that has `SameSite` none, and perhaps with a name that contains the provider name and a short expiration time.


For further checks, read this comment that explains some other considerations with storing authenticated ids and validating if you see them re-used.
https://github.com/simplesamlphp/simplesamlphp/issues/230#issuecomment-123644732

matthewsheerin:
I'm also interested in this, given I'm adding this gem to an app and trying to follow recommended security practices. For example, it's mentioned in the OWASP guide: https://cheatsheetseries.owasp.org/cheatsheets/SAML_Security_Cheat_Sheet.html.

PR #180, opened 2019-11-28
eIDAS SAML samlp:Extensions to AuthRequest option

smarek:
This PR depends on https://github.com/onelogin/ruby-saml/pull/520 being merged and will require binding of omniauth-saml to new version of ruby-saml gem, so I'm pushing the PR for reference and discussion, but don't expect it to be merged before ruby-saml#520 is released

This PR intends to provide samlp:Extensions as per EC eIDAS references ( see https://ec.europa.eu/cefdigital/wiki/display/CEFDIGITAL/eIDAS+eID+Profile )

Basically AuthRequest must contain Extensions with definition of eidas:SPType (ServiceProviderType) and eidas:RequestedAttributes (something that saml-core provides only in SeP metadata)

New option (by default disabled) `:auth_request_include_request_attributes` allows user to configure sending required Extensions in AuthRequest, and uses options `:sptype` and `:request_attributes` to fill the RubySaml::Settings with necessary info

Points for discussion, because this is my first time with your library, and I'm not sure if the implementation follows your expectations/guidelines

 - Would it be possible to extend `:request_attributes` with `:isRequired(bool)` and `:value(anytype)` symbols for each array member, so the user can configure all params expected to be available in eidas:RequestedAttribute ?
 - eIDAS Name is usually URL specifying the namespace of data (such as `'http://eidas.europa.eu/attributes/naturalperson/DateOfBirth'`), currently your `:name` is plain string
    - can user simply replace the `:name => 'email'` with ie. `:name => 'http://eidas.europa.eu/attributes/naturalperson/CurrentGivenName'` ?
    - if not, shall I provide alternative symbol (such as `:request_name`) to fill the RequestedAttribute correctly?
  - Does merging ruby-saml#520 break this library in any way? (my guess would be new default param in RubySaml::Settings initialize, could cause problems, but i'm not sure how to check that https://github.com/onelogin/ruby-saml/pull/520/files#diff-076fc02818002d2ff779f7b6edc76daaR13 ) 

Anyway excuse my ruby skills, i've just started, hope my PR is readable to you

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -15,28 +15,46 @@ module OmniAuth
       option :name_identifier_format, nil
       option :idp_sso_target_url_runtime_params, {}
       option :request_attributes, [
-        { :name => 'email', :name_format => 'urn:oasis:names:tc:SAML:2.0:attrname-format:basic', :friendly_name => 'Email address' },
-        { :name => 'name', :name_format => 'urn:oasis:names:tc:SAML:2.0:attrname-format:basic', :friendly_name => 'Full name' },
-        { :name => 'first_name', :name_format => 'urn:oasis:names:tc:SAML:2.0:attrname-format:basic', :friendly_name => 'Given name' },
-        { :name => 'last_name', :name_format => 'urn:oasis:names:tc:SAML:2.0:attrname-format:basic', :friendly_name => 'Family name' }
+          {:name => 'email', :name_format => 'urn:oasis:names:tc:SAML:2.0:attrname-format:basic', :friendly_name => 'Email address'},
+          {:name => 'name', :name_format => 'urn:oasis:names:tc:SAML:2.0:attrname-format:basic', :friendly_name => 'Full name'},
+          {:name => 'first_name', :name_format => 'urn:oasis:names:tc:SAML:2.0:attrname-format:basic', :friendly_name => 'Given name'},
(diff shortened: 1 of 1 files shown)


smarek:
Travis CI tests fail from obvious reasons, but locally the whole thing can be tested using Bundler git/github repository feature

Step.1: require updated ruby-saml gem from github via Bundler instead of gemspec (see repo diff lower)
Step.2: bundle exec rspec

```diff
> git diff
diff --git a/Gemfile b/Gemfile
index 3cb5947..71e5285 100644
--- a/Gemfile
+++ b/Gemfile
@@ -11,5 +11,6 @@ group :test do
 end
 
 gem 'appraisal'
+gem 'ruby-saml', :github => 'smarek/ruby-saml', :branch => 'eidas-saml-extensions'
 
 gemspec
diff --git a/omniauth-saml.gemspec b/omniauth-saml.gemspec
index 796796f..bc8e989 100644
--- a/omniauth-saml.gemspec
+++ b/omniauth-saml.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |gem|
   gem.required_ruby_version = '>= 2.1'
 
   gem.add_runtime_dependency 'omniauth', '~> 1.3', '>= 1.3.2'
-  gem.add_runtime_dependency 'ruby-saml', '~> 1.8'
+#  gem.add_runtime_dependency 'ruby-saml', '~> 1.11.1'
 
   gem.add_development_dependency 'rake', '>= 10', '< 12'
   gem.add_development_dependency 'rspec', '~>3.4'
```

PR #186, opened 2020-04-08
Rework .travis.yml

bufferoverflow:


bufferoverflow:
Thanks @toupeira I think we can add the `appraisal` any time if needed. I prefer to have only the things we really need.

bufferoverflow:
@md5 @supernova32 Could you review?

PR #195, opened 2020-10-06
fix: Signature validation on IdP initiated SLO

### LINKED ISSUE #194: Signature validation always true with IdP initiated SLO
Hello everyone,
In one project we use Omniauth with SAML.
And we always get valid signature even it should not be valid.
After debug, I figured out that the problem in method `handle_logout_request`

Method:
`def handle_logout_request(raw_request, settings)
        logout_request = OneLogin::RubySaml::SloLogoutrequest.new(raw_request)

        if logout_request.is_valid? &&
          logout_request.name_id == session["saml_uid"]

          # Actually log out this session
          options[:idp_slo_session_destroy].call @env, session

          # Generate a response to the IdP.
          logout_request_id = logout_request.id
          logout_response = OneLogin::RubySaml::SloLogoutresponse.new.create(settings, logout_request_id, nil, RelayState: slo_relay_state)
          redirect(logout_response)
        else
          raise OmniAuth::Strategies::SAML::ValidationError.new("SAML failed to process LogoutRequest")
        end
      end`
It creates **logout_request** object without any settings. But `validate_signature` method use it with two keys "settings" and "get_params"

Maybe someone could explain is it my fault? Or how to fix it?

RomanKosovnenko:
This closes #194

Hello everyone,
In this PR I would like to fix Issue #194 
The solution is to provide a hash with two keys "settings" and "get_params". 

Look forward for your reviews and suggestion, because I'm not a ruby developer.

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -177,7 +177,7 @@ module OmniAuth
       end
 
       def handle_logout_request(raw_request, settings)
-        logout_request = OneLogin::RubySaml::SloLogoutrequest.new(raw_request)
+        logout_request = OneLogin::RubySaml::SloLogoutrequest.new(raw_request, {}.merge(settings: settings).merge(get_params: @request.params))
 
         if logout_request.is_valid? &&
           logout_request.name_id == session["saml_uid"]


bufferoverflow (approved):

bukowczanmarcin00-sketch (commented):
@dennispaagman

PR #216, opened 2024-01-08
Fix copy-pasteability of code example

tjschuck:
Though this code is already not copy-pasteable as valid Ruby (due to the trailing `...` at the end of the `idp_cert_multi` array), that's at least obviously invalid and fast enough to fix.

This missing comma, however, causes parser errors that are less obvious and should be added!


suprnova32 (approved):

bukowczanmarcin00-sketch (commented):
[]()README.md
@nicolasleger 
- __

Issue #229, opened 2024-09-16
Is there going to a minor version patch for OmniAuth 1.0 to address Ruby SAML vulnerability


kli-instacart:
For the CVE issue highlighted by: GHSA-cvp8-5r8g-fhvq, regarding Ruby SAML vulnerability, are there plans to add a minor version bump to address the issue?

Our team relies on OmniAuth 1.0, but do not currently have bandwidth to undertake a major version bump. A minor patch would help bridge the gap.

stephenpowers-sage:
According to dependobot alert, omniauth-saml has a v1.17 patch for this issue https://github.com/SAML-Toolkits/ruby-saml/security/advisories/GHSA-jw9c-mfg7-9rx2 but I see no tag for this. We are in the same position as @kli-instacart

bufferoverflow:
Here you go https://github.com/omniauth/omniauth-saml/releases/tag/v1.10.4

stephenpowers-sage:
Thanks @bufferoverflow much appreciated :)

Issue #228, opened 2024-09-14
Version 2.1.1 breaks semantic versioning


chiragsinghal:
In patch version update from 2.1.0 to 2.1.1, ruby version requirement was modified from 2.4+ to 3.1+, this could be a breaking change for apps running on Ruby 3.0 or lower.
If possible, please consider including only the required security patch related changes in this patch version update instead of modifying the minimum ruby version requirements. Those who are on higher ruby version can use the 2.2.x versions already.

bufferoverflow:
Just fixed that see https://github.com/omniauth/omniauth-saml/releases/tag/v2.1.2 , see https://rubygems.org/gems/omniauth-saml/versions/2.1.2

chiragsinghal:
Thanks a lot for quick action

PR #230, opened 2024-09-17
fix: revert required_ruby_version to >=2.4 for 1.10.x

ilikepi:
In #228, it was noted that the required_ruby_version setting for the gem was bumped between gem version 2.1.0 to 2.1.1.  This was addressed in e9d5238 and released as gem version 2.1.2.

Earlier today, gem version 1.10.4 was released in order to resolve https://github.com/advisories/GHSA-cvp8-5r8g-fhvq for older rubies (ref #229), however this also highlights an earlier case in which the required_ruby_version setting was bumped during a patch release.  In this case, between gem version 1.10.1 and 1.10.2, required_ruby_version was bumped from >=2.1 to >=2.4.

With this PR, I am requesting that the required_ruby_version change made in gem version 1.10.2 be reverted in order to expand the ability for users stuck on the 1.10.x tree to resolve GHSA-cvp8-5r8g-fhvq.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ https://github.com/omniauth/omniauth-saml
 ## Requirements
 
 * [OmniAuth](http://www.omniauth.org/) 1.3+
-* Ruby 2.4.x+
+* Ruby 2.1.x+
 
 ## Versioning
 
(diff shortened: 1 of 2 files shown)


bufferoverflow:
here you go https://rubygems.org/gems/omniauth-saml/versions/1.10.5

ilikepi:
Thank you so much!

PR #232, opened 2025-01-20
feat: log errors on failed logout

machisuji:
This is a mini PR that merely includes actual error messages in logout request validation errors, giving people a chance to find out what's wrong when it's not working.

### DIFF
--- a/lib/omniauth/strategies/saml.rb
+++ b/lib/omniauth/strategies/saml.rb
@@ -190,7 +190,7 @@ module OmniAuth
           logout_response = OneLogin::RubySaml::SloLogoutresponse.new.create(settings, logout_request_id, nil, RelayState: slo_relay_state)
           redirect(logout_response)
         else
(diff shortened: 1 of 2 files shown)


suprnova32 (approved):

jamgregory:
This would be really useful for us - we're currently getting a lot of these exceptions but don't know what is causing them. Any idea when the gem might be bumped to include this @suprnova32 ?

bufferoverflow:
see #233

jamgregory:
I don't think this is quite working as expected - it looks like the SloLogoutrequest instance expects `is_valid?` to be passed the parameter `true` if you want it to collect the errors and return them to the caller:
https://github.com/SAML-Toolkits/ruby-saml/blob/master/lib/onelogin/ruby-saml/slo_logoutrequest.rb#L58

Happy to put in another PR to fix this unless you just want to change it directly :+1:

Issue #236, opened 2025-03-18
Changelog commit links broken in release branches


ilikepi:
Just a heads up that the commit links in CHANGELOG.md are broken in the release branches.  This is because they are relative links starting with "../..".  The release branch naming convention includes a `/` which winds up in the path of the URL when viewing CHANGELOG.md on github.com, so the relative links would need one additional "../" to resolve correctly in Github's URL scheme.  I'm happy to submit a PR for this against the release branches, but I wasn't sure if the change would be overwritten by a script the next time a release was created...

Also note that if you test this yourself, the commit link at the top of CHANGELOG.md in `master` uses an invalid commit hash.  Presumably the original commit was rewritten during the PR that added it.

PR #237, opened 2025-03-30
chore: fix commit hash within changelog

bufferoverflow:
Related #236


fh1ch (approved):
Great catch, I‘ve verified the link. Thanks for fixing this 💚 

LGTM 👍

PR #250, opened 2026-02-17
Re-sign XML test fixtures with SHA-256 instead of SHA-1

gerardo-navarro:
- Migrate all SAML XML fixture signatures from rsa-sha1/sha1 to rsa-sha256/sha256 to fix cross-platform test failures caused by libxml2 canonicalization differences and SHA-1 deprecation.
- Unify custom_attributes.xml to use the same cert/key as other fixtures (was previously using a separate flowdock cert with no matching private key in the repo).
- Update digest_mismatch.xml and invalid_signature.xml to use SHA-256 algorithms while preserving their intentionally corrupted values.
- Remove now-unnecessary idp_cert_fingerprint overrides in spec for custom_attributes contexts.


spec/omniauth/strategies/saml_spec.rb
```diff
@@ -296,7 +296,6 @@
       let(:xml) { :custom_attributes }
 
       before :each do
-        saml_options[:idp_cert_fingerprint] = "3B:82:F1:F5:54:FC:A8:FF:12:B8:4B:B8:16:61:1D:E4:8E:9B:E2:3C"
```
gerardo-navarro: Removing this `idp_cert_fingerprint` override because the `custom_attributes.xml` fixture has been re-signed with the project's shared test certificate (`example_cert.pem`). The fixture was originally added in [commit abdd336](https://github.com/omniauth/omniauth-saml/commit/abdd336a5fb274a5c600d8d7c5aa46823de5dcb6) by a Flowdock contributor, and was likely signed with a certificate from the original contributor whose matching private key was never added to the project. This required a per-context fingerprint override to make the test pass. Now that all fixtures use the same certificate, the default fingerprint (`C1:59:74:2B:...`) defined in `saml_options` applies correctly to all test contexts, making this override unnecessary.

bufferoverflow (approved):

